PR #44, opened 2026-03-11
Fix Soniox STT: Lost Transcripts, Cross-Session Leakage & API Key Handling

razllivan:
## Context

Soniox STT uses a persistent WebSocket connection and a token-based streaming model. Unlike REST-based providers (Groq, OpenAI) that return complete final transcripts per request, Soniox streams individual tokens marked as `is_final: true/false`. The last spoken segment often stays in an **interim** (non-final) state until more audio arrives or the server detects an endpoint.

This streaming model introduced two bugs specific to the **Answer** button flow (manual recording → stop → send to LLM).

---

## Bug 1: "No Speech Detected" After Speaking

**Symptoms:** User clicks Answer, speaks a full sentence, clicks Stop — but sees "⚠️ No speech detected" as if they said nothing.

**Cause:** When the user clicks Stop, `handleAnswerNow` reads the accumulated transcript from React state. However, Soniox often hasn't promoted the last spoken segment to `final` yet — it's still sitting as an interim token. React state updates are asynchronous, and the interim value was captured via `useState` inside a closure that was already stale by the time Stop fires. The result: `voiceInput` is empty or missing the last phrase.

**Fix:** Added `manualTranscriptRef` (a `useRef`) that mirrors the latest interim transcript synchronously. On Stop, the question is assembled as:

```typescript
const question = (voiceInputRef.current + manualTranscriptRef.current).trim();
```

This captures both finalized segments (`voiceInputRef`) and any trailing interim text (`manualTranscriptRef`) at the exact moment of the click — no race condition, no stale closure.

> **Other providers unaffected:** REST-based providers (Groq, OpenAI) only emit `final: true` transcripts, so `manualTranscriptRef` is always empty at Stop time. Google and Deepgram finalize quickly, same situation. The concatenation is a no-op for them.

---

## Bug 2: Previous Answer Leaking Into New Answer

**Symptoms:** User clicks Answer → speaks → Stop → clicks Answer again quickly → the new transcript contains fragments from the previous session.

**Cause:** Soniox maintains audio context on the server across the persistent WebSocket. Clicking Stop disconnected the UI's recording state but did **not** tell the Soniox server that the segment was finished. Residual audio context lingered on the server and produced stale tokens when the next Answer session started.

**Fix — two layers of defense:**

### Layer 1: `enable_endpoint_detection`

Enabled Soniox's built-in endpoint detection in the WebSocket config. This causes the server to auto-finalize segments when it detects a natural speech pause, which significantly reduces the amount of lingering context between sessions.

However, endpoint detection alone is **not sufficient**. If the user clicks Stop and immediately clicks Answer again (before the server detects an endpoint), residual tokens from the old session can still leak into the new one.

### Layer 2: Manual Finalization (`finalize()`)

Added explicit segment finalization per Soniox API docs. When the user clicks Stop, the UI sends a `finalizeMicSTT` IPC call that triggers:

```typescript
this.ws.send(JSON.stringify({ type: 'finalize' }));
```

This forces the Soniox server to immediately flush all pending tokens and emit a `<fin>` marker, guaranteeing a clean slate for the next session — regardless of timing.

The `<fin>` marker token (and `<end>` from endpoint detection) are silently filtered in the message handler to prevent them from appearing in the transcript UI.

---

## Additional Fixes

- **`config.languages` → `config.language_hints`** — corrected the Soniox API config key per their [documentation](https://soniox.com/docs/stt/concepts/language-hints).
- **Simplified token emission** — removed redundant `pendingFinalText` accumulation inside `SonioxStreamingSTT`. The UI already accumulates final segments via `voiceInputRef`, so double-buffering was unnecessary.
- **`STTProvider` type** — introduced a unified type alias with optional `finalize()` and `setAudioChannelCount()` to eliminate `as any` casts in `main.ts`.
- **Settings: Soniox key handling** — fixed a bug where deleting the Soniox API key also deleted the Deepgram key (incorrect fallthrough logic). Added `soniox` to the key submission map.

---

## Changed Files

| File | Summary |
|------|---------|
| `electron/audio/SonioxStreamingSTT.ts` | `finalize()` method, `<fin>`/`<end>` token filtering, `language_hints`, `enable_endpoint_detection`, simplified token emission |
| `electron/main.ts` | `finalizeMicSTT()`, `STTProvider` type alias, removed `as any` casts |
| `electron/ipcHandlers.ts` | New `finalize-mic-stt` IPC handler |
| `electron/preload.ts` | `finalizeMicSTT` bridge to renderer |
| `src/types/electron.d.ts` | `finalizeMicSTT` type declaration |
| `src/components/NativelyInterface.tsx` | `manualTranscriptRef`, finalize call on Stop, interim capture |
| `src/components/SettingsOverlay.tsx` | Soniox key deletion fix, key submit map |

## Compatibility

All changes are backward-compatible with existing STT providers:

- `finalizeMicSTT()` uses optional chaining — only fires for providers that implement `finalize()` (currently only Soniox).
- `manualTranscriptRef` concatenation is a no-op for providers that emit `final: true` promptly.
- No changes to `GoogleSTT`, `DeepgramStreamingSTT`, or `RestSTT` source code.

## Testing

- [x] Soniox: Stop immediately after speaking → full text captured
- [x] Soniox: Rapid Answer → Stop → Answer → Stop → no cross-session leakage
- [x] Soniox: `<fin>` and `<end>` markers do not appear in UI
- [x] Groq STT: No regression
- [x] Deepgram STT: No regression
- [x] "No speech detected" still shows when user says nothing
- [x] TypeScript: `tsc --noEmit` passes with 0 errors

### DIFF
--- a/src/components/NativelyInterface.tsx
+++ b/src/components/NativelyInterface.tsx
@@ -68,6 +68,7 @@ const NativelyInterface: React.FC<NativelyInterfaceProps> = ({ onEndMeeting }) =
     const [isManualRecording, setIsManualRecording] = useState(false);
     const isRecordingRef = useRef(false);  // Ref to track recording state (avoids stale closure)
     const [manualTranscript, setManualTranscript] = useState('');
+    const manualTranscriptRef = useRef<string>('');
     const [showTranscript, setShowTranscript] = useState(() => {
         const stored = localStorage.getItem('natively_interviewer_transcript');
         return stored !== 'false';
@@ -303,9 +304,11 @@ const NativelyInterface: React.FC<NativelyInterfaceProps> = ({ onEndMeeting }) =
                         return updated;
                     });
                     setManualTranscript('');  // Clear partial preview
+                    manualTranscriptRef.current = '';
                 } else {
                     // Show live partial transcript
                     setManualTranscript(transcript.text);
+                    manualTranscriptRef.current = transcript.text;
                 }
                 return;  // Don't add to messages while recording
             }
@@ -843,12 +846,17 @@ const NativelyInterface: React.FC<NativelyInterfaceProps> = ({ onEndMeeting }) =
             setIsManualRecording(false);
             setManualTranscript('');  // Clear live preview
 
+            // Send manual finalization signal to STT Providers
+            window.electronAPI.finalizeMicSTT().catch(err => console.error('[NativelyInterface] Failed to send finalizeMicSTT:', err));
+
             const currentAttachment = attachedContext;
             setAttachedContext(null); // Clear context immediately on send
 
-            const question = voiceInputRef.current.trim();
+            const question = (voiceInputRef.current + (manualTranscriptRef.current ? ' ' + manualTranscriptRef.current : '')).trim();
             setVoiceInput('');
             voiceInputRef.current = '';
+            setManualTranscript('');
+            manualTranscriptRef.current = '';
 
             if (!question && !currentAttachment) {
                 // No voice input and no image


evinjohnn:
The issue is that I’m using a 256 GB Mac, so I’m not able to run VMware to develop the Windows version locally. Because of that, I currently have to wait until I can use a friend’s Windows laptop to work on it.

If you could send me the release files, I can upload them to the Releases section myself. That would make them available sooner and help a lot of users.

Issue #139, opened 2026-03-29
[Feature Request]: Some important features that are currently missing


caluclatedmoment-arch:
### Description

1. Allow the user to customize the prompt, or upload documents instead of just resume to provide personalized answers. for example, I want to provide a bunch of leetcode notes, and let the ai retrieve information from these notes to generate a solution that I am most familiar with.
2. Adjust the size of the copilot window automatically or allow manual adjustments so that all lines of code can be fully displayed.
3. Now when an answer is being generated, the UI forces alignment to the last line of answer generated, meaning that we have to wait for the answer to finish generating before we can start reading it. It would be nice to keep the UI where it was, so that we can see the answer from the beginning, and don't have to manually scroll up to the first line of the answer.
4. Now we need to navigate 3 times and click 3 times to delete a single meeting, which is pretty troublesome, it will be much more convenient to add a select to delete feature.
5. It would be nice to also keep listening and transcribing the speech of the user, so that the user can use a trigger sentence/words, eg. "let me think", "good question" to generate the answer automatically.

Thank you for your great work, hopefully these features can be added in the incoming updates.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

evinjohnn:
<img width="1194" height="798" alt="Image" src="https://github.com/user-attachments/assets/f587b680-0de0-44d8-ad14-7ab2ef219cb5" />

<img width="1201" height="796" alt="Image" src="https://github.com/user-attachments/assets/aea3ac4a-e857-4520-98f5-697980e85e23" /> yes i have added modes option and custom context option also , would be available in next update

PR #150, opened 2026-04-02
fix: improve custom provider streaming and chat auto-scroll behavior

WLAXID:
## Summary

Fixes two issues: custom provider streaming not working at all, and chat auto-scroll interfering with message reading during streaming responses (point 3 in #139). 

**Problems fixed:**

- Custom provider streaming completely broken with error: `[Error: Unexpected token 'd', "data: {"id"... is not valid JSON]`
- Custom providers were incorrectly treated as `CurlProvider` instead of using SSE-capable `streamWithCustom()` method
- Streaming responses showed JSON artifacts like `{"choices":[{"delta":{}}]}` in chat output
- Auto-scroll triggered on every message update, making it impossible to read earlier messages while AI was streaming

### Custom Provider Streaming Fix

**electron/LLMHelper.ts** - Fixed provider switching logic in `switchToModel()`:

- `customProvider` was being reset to `null` and provider was incorrectly assigned to `activeCurlProvider`
- Now correctly sets `customProvider` and keeps `activeCurlProvider` as `null`
- Reordered streaming handlers: CustomProvider check now comes before CurlProvider fallback

**electron/LLMHelper.ts** - Improved `extractFromCommonFormats()` for streaming responses:

- Returns empty string for delta chunks with no extractable content (prevents JSON artifacts)
- Properly handles empty choices arrays (usage/finish chunks)
- Added comment explaining why `reasoning_content` is intentionally not extracted
- Added guard to skip SSE-formatted data when parsing as JSON fallback

**electron/LLMHelper.ts** - Fixed `streamWithCustom()` SSE parsing:

- Added check to skip JSON parsing if body starts with `"data: "` (SSE format)

### Chat Auto-Scroll Behavior

**src/components/GlobalChatOverlay.tsx** - Removed auto-scroll on every message update, added scroll only when user sends message (50ms delay for DOM update)

**src/components/NativelyInterface.tsx** - Same fix: removed aggressive auto-scroll, added scroll-to-bottom on message send

**src/components/MeetingChatOverlay.tsx** - Same fix applied

## Type of Change

- [x] 🐛 Bug Fix
- [ ] ✨ New Feature
- [ ] 📝 Documentation

## Testing & Environment

- [x] Manual test performed on: **Windows 11**

**How to verify:**

1. Configure a custom OpenAI-compatible provider in Settings
2. Send a message using the custom provider - verify streaming works without JSON artifacts
3. While AI is streaming a response, scroll up to read earlier messages - verify view stays in place
4. Send another message - verify chat scrolls to bottom correctly

### Custom Provider Issue

The `switchToModel()` function had incorrect logic:

```typescript
// BEFORE (broken):
this.customProvider = null;
this.activeCurlProvider = custom; // Treats custom provider as non-streaming cURL

// AFTER (fixed):
this.customProvider = custom;
this.activeCurlProvider = null; // Uses streamWithCustom() for SSE streaming
```

This caused custom providers to bypass the proper SSE streaming implementation and use the non-streaming cURL fallback instead. The cURL fallback tried to parse SSE-formatted response as JSON, resulting in:

```
[Error: Unexpected token 'd', "data: {"id"... is not valid JSON]
```

### Streaming Response Parsing

The `extractFromCommonFormats()` function didn't handle streaming delta chunks without content:

- Delta chunks like `{"choices":[{"delta":{}}]}` were stringified and shown to users
- Empty choices arrays (usage stats) were also stringified

### Auto-Scroll Issue

The `useEffect` with `messages` dependency triggered scroll on every message update:

```typescript
// BEFORE: triggers on every streaming chunk
useEffect(() => {
  messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
}, [messages]);

// AFTER: only scrolls when user sends message
const handleSend = () => {
  // ... add message ...
  setTimeout(() => {
    messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
  }, 50);
};
```

### DIFF
--- a/electron/LLMHelper.ts
+++ b/electron/LLMHelper.ts
@@ -232,10 +232,9 @@ export class LLMHelper {
     const custom = customProviders.find(p => p.id === targetModelId);
     if (custom) {
       this.useOllama = false;
-      this.customProvider = null;
-      // Treat text-only custom providers as CurlProviders (responsePath optional)
-      this.activeCurlProvider = custom as CurlProvider;
-      console.log(`[LLMHelper] Switched to cURL Provider: ${custom.name}`);
+      this.customProvider = custom;
+      this.activeCurlProvider = null;
+      console.log(`[LLMHelper] Switched to Custom Provider: ${custom.name}`);
       return;
     }
 
@@ -1378,6 +1377,9 @@ This rule overrides ALL other instructions including formatting, brevity, or out
     // OpenAI delta/streaming format: { choices: [{ delta: { content: "..." } }] }
     if (data.choices?.[0]?.delta?.content) return data.choices[0].delta.content;
 
+    // NOTE: reasoning_content (model's thinking process) is intentionally NOT extracted
+    // to avoid showing internal reasoning to users. Only final content is returned.
+
     // Anthropic format: { content: [{ text: "..." }] }
     if (Array.isArray(data.content) && data.content[0]?.text) return data.content[0].text;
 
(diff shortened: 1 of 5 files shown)


evinjohnn:
> @evinjohnn - sorry to nag, but could you take a look at the streaming issue? It’s becoming a bit of a blocker, so any help with a quick fix would be much appreciated

hey i didnt understand the issue , can you explain it a bit more

PR #320, opened 2026-06-17
fix(llm): treat claude-opus-4-8 as Opus for prompt-cache min-chars

sumleo:
## What

`electron/LLMHelper.ts` — `getClaudeCacheMinChars` (the per-model minimum prompt size for Anthropic prompt caching to engage) enumerated Opus point releases by exact prefix:

```ts
// electron/LLMHelper.ts (before)
if (id.startsWith("claude-opus-4-7") || id.startsWith("claude-opus-4-6") || id.startsWith("claude-opus-4-5") || id.startsWith("claude-haiku-4-5")) return 4096 * 4;
```

`claude-opus-4-8` is not in that list, so it fell through to the generic `claude-` branch and returned the **1,024-token (4,096-char)** floor instead of the **4,096-token (16,384-char)** minimum that every Opus 4.5+ model requires. For prompts whose size lands between those two thresholds, Anthropic silently skips prompt caching — the request still succeeds, `cache_creation_input_tokens` stays `0`, and full input price is paid on every turn.

## Fix

Match on the `claude-opus-4-` family prefix instead of enumerating point releases, mirroring the sibling `getClaudeMaxOutput`, which already uses the robust prefix:

```ts
// getClaudeMaxOutput (existing, same file)
if (id.startsWith("claude-opus-4-0") || id.startsWith("claude-opus-4-1")) return 32000;
if (id.startsWith("claude-opus-4-")) return 128000;
```

So `getClaudeCacheMinChars` now does the same — Opus 4.0 / 4.1 keep the 1,024-token floor (they predate the bump), and every Opus 4.5 and later (4.5 / 4.6 / 4.7 / 4.8, plus future point releases) gets the correct 4,096-token floor. The `claude-haiku-4-5` check is preserved.

## Changes

- `electron/LLMHelper.ts`: prefix-match the Opus family; update the threshold docstring to mention Opus 4.8.
- `electron/llm/__tests__/ClaudeCacheMinChars.test.mjs`: new `node:test` asserting `claude-opus-4-8 → 16,384` chars and pinning the other per-model floors so the regression can't return. Exercises the real compiled `LLMHelper` method (same private-method access pattern as `NegotiationStickinessAndCircuitBreaker.test.mjs`).
- `CHANGELOG.md`: one-line entry under `## [Unreleased]`.

## Testing

- `npm run typecheck:electron` — clean.
- `npm run build:electron` then `node --test electron/llm/__tests__/ClaudeCacheMinChars.test.mjs` — 9/9 pass. The `claude-opus-4-8 → 16,384` assertion fails against the pre-fix code (it returned 4,096) and passes after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
     - **LiteLLM AI Gateway**: Added LiteLLM as a built-in provider, giving access to 100+ LLM providers (AWS Bedrock, Google Vertex AI, Azure, Cohere, and more) through a single OpenAI-compatible proxy. Configure the proxy URL and optional virtual key under Settings → AI Providers → LiteLLM Proxy; models are auto-discovered from the proxy and listed with a `litellm/` prefix. Max output tokens default to **Auto** — each model's real output budget is read from the proxy's `/model/info` registry (fallback 8,192) — with a manual dropdown override (4K–1M). Routes through the same data-scope gating, rate-limiting, and abort-aware streaming as every other cloud provider.
 
+    ### Improvements & Fixes
+
+    - **Prompt caching for Claude Opus 4.8**: `getClaudeCacheMinChars` now matches the whole `claude-opus-4-` family instead of enumerating point releases, so `claude-opus-4-8` uses the correct 4,096-token (16,384-char) cache minimum. It previously fell through to the generic 1,024-token floor, which silently disabled prompt caching for prompts between those two sizes.
+
     ## [2.7.0] - 2026-06-05
 
     ### What's New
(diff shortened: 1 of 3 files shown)


sumleo:
Thanks for the review. Addressed the inline point in 8fe6a9e: the Opus 4.0/4.1 carve-out now anchors on a terminal version digit (`/^claude-opus-4-[01](-|$)/`), so a future two-digit point release like `claude-opus-4-10` is no longer captured by the `claude-opus-4-1` prefix and correctly gets the 4,096-token floor. Added tests for the two-digit case and for dated 4.0/4.1 snapshot ids.

On the CHANGELOG note: I kept the `### Improvements & Fixes` subsection because it matches the structure used in the released sections (e.g. 2.7.0 has both `### What's New` and `### Improvements & Fixes`) — a fix belongs under the latter rather than under `What's New`.

evinjohnn:
Merged directly into main at 024f365. Thanks!

sumleo:
Hi @evinjohnn, gentle nudge on this when you have a moment. It's a small, self-contained prompt-caching fix, and I'm happy to rebase or tweak anything if that would make review easier. Thanks for the project and your time!

PR #364, opened 2026-07-11
fix: add visible close button to settings panel

### LINKED ISSUE #363: Settings panel has no visible close button — sidebar layout pushes it out of view
﻿## Description

The Settings panel (SettingsOverlay.tsx) lacks a convenient way to close it. There is a Close button at the bottom of the sidebar, but it is **invisible** because of a CSS flex layout issue.

## Root Cause

The sidebar structure has a flex layout problem. The nav section uses lex-1 which consumes all available vertical space (12 nav items fill the entire sidebar height). The bottom div with the Close button uses mt-auto but since the nav div already fills 100% of the container, the bottom div is pushed below the visible viewport.

The Close button exists in the DOM but is scrolled out of sight. Users have to rely on pressing Escape or clicking the backdrop to close settings — there is no visible affordance.

## Workarounds
- Press Escape key to close settings
- Click outside the settings modal

## Suggested Fix
Add a prominent close button (X icon) in the top-right corner of the settings panel, which is always visible regardless of sidebar content height. The sidebar layout should also be fixed so the existing bottom Close button is reachable.

rafifshaf-fun:
﻿## Description

Adds a visible close button to the Settings panel. The existing Close button at the bottom of the sidebar was invisible because the sidebar's flex layout (flex-1 on the nav section) consumed all available height, pushing the bottom div below the viewport.

## Changes

- Added a prominent X close button in the top-right corner of the settings panel wrapper (absolute position, high z-index, always visible)
- Button matches the app's design language with backdrop blur and hover effects

## Before

Users had to press Escape or click the backdrop to close settings — no visible affordance existed.

## After

A clear X button in the top-right corner is always visible regardless of sidebar content height. The existing Close button at the bottom of the sidebar is also preserved.

## Testing

- npm run typecheck:electron passes
- npm run build (Vite renderer) succeeds
- npm run build:electron succeeds

Closes #363

### DIFF
--- a/electron/WindowHelper.ts
+++ b/electron/WindowHelper.ts
@@ -438,9 +438,21 @@ export class WindowHelper {
     // line naming the exact failing module/line on the user's machine.
     this.attachRendererDiagnostics(this.launcherWindow, 'launcher');
 
-    // if (isDev) {
-    //   this.launcherWindow.webContents.openDevTools({ mode: 'detach' }); // DEBUG: Open DevTools
-    // }
+    // DIAGNOSTIC (2026-07-11): NATIVELY_OPEN_DEVTOOLS=1 force-opens the launcher
+    // DevTools DETACHED (survives a renderer hang, unlike an in-window panel).
+    // For debugging the "window appears then freezes / renderer not responsive"
+    // report: open the Performance tab and record during the freeze — a single
+    // long yellow Scripting block = a JS main-thread loop; a flat gap with no JS
+    // = a compositor/GPU stall (the software-compositing blur path). Detached so
+    // it stays usable even when the launcher renderer stops pumping its loop.
+    if (process.env.NATIVELY_OPEN_DEVTOOLS === '1') {
+      try {
+        this.launcherWindow.webContents.openDevTools({ mode: 'detach' });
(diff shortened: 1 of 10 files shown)


evinjohnn:
Merged the UI fix in commit [`931e447`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/commit/931e447).

I had to omit two things from this PR:

1. **`premium` and `natively-api` submodule pointer moves.** Both were going *backwards* to older snapshots (premium `8b9f91b` → `9e04f09`, natively-api `dd547cd` → `0c51f01`). Per repo review rules, submodule pointer bumps need to be explicitly justified and tied to Pro-tier integration — a 10-line UI fix isn't. Resetting to origin/main's values keeps Pro-tier code current.

2. **Sidebar layout + STT broadcast-order fixes mentioned in the commit body.** Grep on `flex-1`, `sidebar`, `reconfigureSttProvider`, `credentials-changed` against the diff returned zero matches. The commit message references fixes that aren't in the code. Future `git log -S` searches for the broadcast-order fix would turn up nothing here.

If either of those two items is real work you intended to ship, please open them as separate PRs (or follow-up commits) with the actual diff and a clearer commit body. The UI fix itself is solid — the opacity-preview visibility gating matches `#settings-panel` exactly.

Closing this PR as superseded — the fix is on `main`, and the other claims need their own review.

PR #370, opened 2026-07-13
fix: fall back to app-managed encryption when OS keyring save fails

### LINKED ISSUE #369: [Bug]:
### What happened?

A newly created permanent Deepgram API key works successfully against the official GET /v1/models endpoint using Authorization: Token <key>. It also initially passes Natively’s Test Connection. However, Natively later returns 401 INVALID_AUTH, sometimes after several seconds or after restarting the application. Repasting the exact same valid key temporarily fixes it. This suggests Natively is corrupting, overwriting, or incorrectly retrieving the stored credential.

### Steps to Reproduce

1. Open Natively.
2. Go to Settings → Audio → Speech Provider.
3. Select Deepgram Nova-3.
4. Enter a newly created permanent Deepgram API key.
5. Click Save.
6. Click Test Connection and confirm that it succeeds.
7. Wait approximately 15 seconds, or fully close and reopen Natively.
8. Click Test Connection again or attempt to start transcription.
9. Observe the following error:

   `Unexpected server response: 401 - {"err_code":"INVALID_AUTH","err_msg":"Invalid credentials.","request_id":"019f5990-aa76-7d53-a2ca-a90ec21c5919"}`

10. Delete the saved key, paste the exact same key again, and click Test Connection.
11. Observe that the connection succeeds again temporarily.

The same API key remains valid when tested directly against Deepgram's official `/v1/models` endpoint using the `Authorization: Token <API_KEY>` header.

### Environment

- OS: Windows 10
- App version: v2.7

### Relevant Log Output

```shell
Unexpected server response: 401 - {"err_code":"INVALID_AUTH","err_msg":"Invalid credentials.","request_id":"019f5990-aa76-7d53-a2ca-a90ec21c5919"}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

rafifshaf-fun:
﻿## Description

Fixes a bug where STT API keys (Deepgram, Groq, OpenAI, etc.) appeared to save successfully but were silently lost after restarting the app.

## Root Cause

\CredentialsManager.saveCredentials()\ prefers the OS keyring (\safeStorage\ via DPAPI on Windows). When \isEncryptionAvailable()\ returns \	rue\, it writes via \safeStorage.encryptString()\ and then **deletes the fallback file**. However, on Windows, DPAPI can throw after \isEncryptionAvailable()\ returns \	rue\ — for example due to policy restrictions, roaming profiles, or sandboxing regressions.

The old code placed BOTH the keyring write AND the fallback write inside the same \	ry\ block. When the keyring write threw, execution jumped straight to the \catch\, returning \alse\ — **without ever attempting the fallback**. The key was in memory (so it worked during the session) but never reached disk, so it vanished after restart.

## Fix

- The keyring write is now wrapped in its own \	ry-catch\
- If it throws, it logs a warning and falls through to the app-managed encrypted fallback (AES-256-GCM via \credentialFallbackCrypto.ts\)
- The fallback path now has its own separate \	ry-catch\ for genuinely unwritable disk scenarios

## Testing

- \
pm run typecheck:electron\ passes
- \
pm run build:electron\ succeeds

Closes #369

### DIFF
--- a/electron/WindowHelper.ts
+++ b/electron/WindowHelper.ts
@@ -438,9 +438,21 @@ export class WindowHelper {
     // line naming the exact failing module/line on the user's machine.
     this.attachRendererDiagnostics(this.launcherWindow, 'launcher');
 
-    // if (isDev) {
-    //   this.launcherWindow.webContents.openDevTools({ mode: 'detach' }); // DEBUG: Open DevTools
-    // }
+    // DIAGNOSTIC (2026-07-11): NATIVELY_OPEN_DEVTOOLS=1 force-opens the launcher
+    // DevTools DETACHED (survives a renderer hang, unlike an in-window panel).
+    // For debugging the "window appears then freezes / renderer not responsive"
+    // report: open the Performance tab and record during the freeze — a single
+    // long yellow Scripting block = a JS main-thread loop; a flat gap with no JS
+    // = a compositor/GPU stall (the software-compositing blur path). Detached so
+    // it stays usable even when the launcher renderer stops pumping its loop.
+    if (process.env.NATIVELY_OPEN_DEVTOOLS === '1') {
+      try {
+        this.launcherWindow.webContents.openDevTools({ mode: 'detach' });
+        console.warn('[Diag] NATIVELY_OPEN_DEVTOOLS=1 → launcher DevTools opened (detached)');
+      } catch (e: any) {
+        console.warn('[Diag] openDevTools failed:', e?.message || e);
+      }
+    }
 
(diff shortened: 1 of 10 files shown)


evinjohnn:
Superseded by #371 (which will be opened shortly). PR #370 was based on a stale fork (parent addc42c2 from 2026-07-11, current main has 4 days of changes). Closing in favor of #371 which contains PR #370's fix PLUS senior-review follow-ups (load-time stale-keyring defense, error message redaction, log phrasing, deduped try/catch, comment expansion) and 4 new regression tests.

PR #381, opened 2026-07-17
fix: make Local Whisper shutdown cooperative

Lob0Garou:
## Summary

- shut down Local Whisper workers cooperatively instead of force-terminating ONNX inference on Windows
- serialize worker teardown and keep the shared ONNX slot until the worker has actually exited
- prioritize the visible interviewer/system channel for the preloaded warm worker
- wait for pending meeting teardown before starting another audio test

## Root cause

`Worker.terminate()` could interrupt ONNX Runtime while native inference was still active. On Windows this could abort the Electron process or leave native resources unwinding while a new worker started. The previous lifecycle also released the shared ONNX slot on the worker `error` event, before the corresponding `exit` confirmed teardown.

## Safety properties

- the worker stops accepting new work after a shutdown request
- already-started operations finish before the pipeline is disposed
- the message port closes only after disposal completes
- a replacement worker cannot overlap the previous native ONNX session
- the ONNX slot is released exclusively from the worker exit path
- meeting and audio-test transitions wait for the teardown barrier
- concurrent audio-test requests acquire their guard before awaiting teardown
- a wedged worker times out after 15 seconds without releasing its ONNX slot; meeting persistence continues and a later Local Whisper start can fail over safely

## Validation

- `npm run typecheck:electron`
- `npm run build:electron`
- 11 focused lifecycle and ordering tests pass
- `git diff --check`
- staged-diff scan found no credentials, local user paths, or sandbox overrides

## Baseline note

A broader Windows audio-test subset still contains three pre-existing failures caused by Node rejecting absolute `C:` ESM imports (`ERR_UNSUPPORTED_ESM_URL_SCHEME`). The focused tests covering this change pass, and this PR does not modify that test-loader behavior.

## Product impact

This addresses the crash/freeze path seen after stopping and restarting Local Whisper capture, while also reducing the cold-start delay for the transcript users actually see during meetings.


Lob0Garou:
Addressed the three lifecycle findings in 7bae3b4: shutdown-send failure now waits for the authoritative worker exit, the audio-test guard is acquired before teardown awaits, and worker waits are bounded without releasing the ONNX slot. Meeting persistence continues after a timeout. Added four regression tests; 11 focused tests, Electron typecheck, and Electron build pass.

PR #388, opened 2026-07-26
fix(context-os): stop withholding JD/resume evidence in unknown-owner turns + Intelligence-OS audit fixes

evinjohnn:
## Scope note (read first)

This branch is **719 commits ahead of `origin/main`** (40 files differ in the
aggregate diff: 2696 insertions / 3124 deletions). This local `main` has been
diverging from `origin/main` for a while; the top commit (`83e049da`) is this
session's actual deliverable. Everything below it is pre-existing local
history that was never previously pushed/PR'd. Flagging this plainly rather
than presenting it as a small, isolated diff — reconciling the rest of the
divergence is a separate decision for whoever reviews this.

## This session's commit (`83e049da`)

**fix(context-os): stop withholding JD/resume evidence in unknown-owner turns**

- `sourceOwnership.ts`'s canonical-decision branch only granted `profileAllowed`
  on `outcome === 'explicit_granted'`, so the common `'default'` outcome
  (unknown owner, no explicit source switch) silently dropped all profile/JD
  evidence from the prompt even when it was available. Mirrors the
  already-correct `wtaDecisionAllowsCandidateProfile` pattern in
  `IntelligenceEngine.ts`.
- Threads `hasJobDescription` through `SourceAuthorityKernel`/`integration`/
  `ipcHandlers` so the unknown-owner branch can grant a `profile_jd`
  capability, and fixes `turnSourceDecision.ts`'s `defaultDecision()` to
  populate `allowed` for the unknown-owner fallback the same way the
  `profile_only` branch does.
- Adds a `trace`-flag-gated diagnostic log in `LLMHelper.stream` to make
  future context-starvation symptoms distinguishable from provider-side
  failures.
- Fixes 10 stale tests (`ProfileIdentityBaseline`, `IntelligenceOsE2E`) that
  asserted the pre-`ed827532` contract (synthesized final-prose strings)
  against code that now deliberately returns deterministic evidence
  selections for JIT rendering — confirmed via `ProfileTreeService`'s own
  docstring and the compiled module's actual return shape. Test-contract
  drift after an already-shipped architecture change, not a regression.

## Test plan
- [x] New/updated tests for the evidence-withholding fix confirmed RED→GREEN
      via isolated `git stash` on the affected files.
- [x] The two stale-test files pass against the current, documented contract
      (11/11 and 27/27).
- [x] `tsc --noEmit` and `build:electron` clean on this branch's own HEAD.


evinjohnn:
Closing: this branch carries 719 commits of unrelated local history and conflicts with 919 commits on main across 13 files, including a deep two-independent-architecture conflict in NativelyInterface.tsx (streaming-height buffer + deterministic-reveal pacer vs stream-ID superseded-stream guards). Reconciling that divergence is separate work, not something to resolve inline here. Reopening the JD-evidence fix as a smaller, targeted PR if it applies cleanly on its own.

Issue #322, opened 2026-06-17
[Bug]: STT reconnecting


itsmart24:
### What happened?

Audio Capture Issue
System audio is being captured but every sample is silent. This usually means macOS Screen Recording permission needs to be re-granted to this build of Natively. Open System Settings → Privacy & Security → Screen Recording, toggle Natively off and back on, then restart the app. (If you recently rebuilt or updated, the previous grant may not apply.)

### Steps to Reproduce

1. Open Natively Pro on macOS.
2. Start using the app with audio/STT enabled.
3. Notice that the top status shows:

### Environment

* OS: macOS
* App: Natively Pro
* Issue type: STT reconnecting / system audio capture / Screen Recording permission

### Relevant Log Output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Hassansidiqui:
The reason is that when I paste the API key in the audio settings and connect it, it works. But after saving, closing the application, and opening it again, the API key in the audio settings shows as invalid.

surya-edict:
> The reason is that when I paste the API key in the audio settings and connect it, it works. But after saving, closing the application, and opening it again, the API key in the audio settings shows as invalid.

same happening here

AniketSarrin:
Facing something similar on windows. No audio being captured at all.

thewitanowski:
Also getting this.  Key doesn't persist, STT doesn't connect as a result, not connecting to audio... product is basically unusable.

sudhanshuy17:
Facing the same issue when I've saved the api key in the audio section, but when I went over the meeting, it says no audio detected, and when I came back to the audio settings, it's showing the api key is invalid

abhayjb:
this issue is super annoying. Always said STT cannot be connected and then  'Privacy & Security' > 'Screen & System Audio Recording' always kept breaking and I had to restart multiple times. Everytime the API keys saved were erazed causing the STT issue again - the loop was never ending!

evinjohnn:
Fixed — please update to **v2.8.8**. This thread ended up covering three different problems, so taking them separately.

## 1. Key not persisting → STT can't connect (what most of you actually hit)

@Hassansidiqui @surya-edict @thewitanowski @sudhanshuy17 — this is your strand, and it's the most cleanly fixed one. Sorry it took this long.

[...]

PR #397, opened 2026-07-27
fix(settings): persist API keys and clarify STT status (#322)

bhargavap21:
## Summary
- Autosave API keys on input/blur/unmount across Settings (BYOK, Natively, LiteLLM, Tavily, STT)
- Harden CredentialsManager against macOS keychain decrypt failures that wiped keys after restart (#322): clobber guard, fallback fall-through, re-entry banner, `keychain-access-groups` entitlement
- Separate STT UI states: listening (sky) vs reconnect (amber) vs credential-unreadable (key banner) so silent capture no longer looks like reconnect

## Test plan
- [ ] Paste a Groq/Deepgram STT key in Settings → Audio, wait ~1s or blur — restart app — key still Saved / Test Connection works
- [ ] Force undecryptable keyring path (or review Issue322KeychainPersistence tests): empty save must not wipe `credentials.enc`; 3 failures show re-enter banner
- [ ] Confirm `build/entitlements.mac.plist` includes `BJM29W3UQ6.com.electron.meeting-notes` keychain group (signed build)
- [ ] During meeting: awaiting-audio shows as listening (not amber reconnect); credential-unreadable shows key banner + Open Audio Settings
- [ ] `node --test electron/services/__tests__/Issue322KeychainPersistence.test.mjs electron/services/__tests__/KeychainEntitlement.test.mjs` (after `build:electron` / esbuild of CredentialsManager)

Made with [Cursor](https://cursor.com)


bhargavap21:
Addressed Greptile P1/P2 in `171b245`:

- **P1:** Sparse/partial persists over an undecryptable keyring are refused; permanent re-entry stages to fallback only; healed launches merge fallback into the recovered keyring so sibling keys survive.
- **P2:** `useApiKeyAutoSave` unmount flush goes through `runSave()` / `inFlightRef`.

Regression coverage added in `Issue322KeychainPersistence.test.mjs` and `useApiKeyAutoSave.test.mjs`.

bhargavap21:
Addressed the follow-up Greptile findings:

- **Sparse fallback migrate-up:** permanent re-entry writes a `credentials.fallback.staging` marker; while that marker is present, migrate-up will not overwrite `credentials.enc`. Healed launches still merge fallback → keyring and clear the marker.
- **Trailing autosave:** `useApiKeyAutoSave` now queues a follow-up `runSave()` when edits arrive during an in-flight save, so the latest value is persisted.

PR #400, opened 2026-07-27
Pin custom context up to its allocated 600-token budget

### LINKED ISSUE #399: [Bug]: Custom Context silently dropped — pinned block truncates at 1,200 chars and the documented retrieval fallback returns nothing
### What happened?

Custom Context silently fails to reach the model on most turns. Asking a natural question about content that is explicitly in my Custom Context returns "I don't have enough context from the conversation to answer that" / "That wasn't discussed as far as I can tell."

Tracing it through the code, there are two separate mechanisms, and the second one invalidates the documented safety net for the first.

**1. The always-pinned block truncates at 1,200 characters — half of its own token budget.**

`electron/services/ModesManager.ts:1169`

```ts
private static readonly PINNED_INSTRUCTIONS_MAX_CHARS = 1_200;
```

The Custom Context field in the UI accepts 4,000 characters (the counter reads `0/4000`), so anything past 1,200 is replaced with `…[truncated]` before it reaches the prompt.

Worth noting this cap is also well under the budget the architecture already allocates to this layer. `electron/llm/contextRoute.ts:55`:

```ts
custom_context: 600,   // tokens
```

600 tokens is roughly 2,400 characters. The pinned path emits ~300 tokens, leaving half its own allocation unused.

**2. The documented fallback for the truncated remainder returns nothing.**

The comment above the constant says:

> `// Anything longer remains fully available to RETRIEVAL (reference-file path), so nothing is lost.`

For a user who has Custom Context but no uploaded reference file, that is not the case. Custom context is pushed through the same lexical relevance gate as reference files (`ModeContextRetriever.ts`), scored by

```
score = matches / sqrt(queryWords.size × uniqueChunkWords)     // line 449
MIN_RELEVANCE_SCORE = 0.18                                     // line 120
```

and when nothing clears the floor the retriever returns an empty context (line 1550):

```ts
return { snippets: [], formattedContext: '', usedFallback: true };
```

I ported that scoring path to a standalone script and ran a realistic 4,000-char Custom Context (five prose paragraphs about a project) against the question `what did we work on this sprint?`:

| Scenario | queryWords.size | threshold | chunk scores | chunks passed |
|---|---|---|---|---|
| Typed before a call (no transcript) | 5 | 0.1800 | 0.1227, 0.1729 | **0 of 2** |
| Mid-meeting (live transcript) | 100 | 0.1800 | 0.1098, 0.1546 | **0 of 2** |

Two structural reasons a natural question cannot clear the floor here:

- On the non-document-grounded path, the **entire live transcript is concatenated into the retrieval query** (`ModeContextRetriever.ts:918`). That inflates `queryWords.size`, which inflates the `sqrt(querySize × chunkSize)` denominator — so scores fall as a meeting gets longer.
- Both existing rescue paths skip this configuration. The adaptive floor reduction is gated to `!hasTranscript` (line ~998), and the broad-query rescue that drops the threshold to 0 requires `forceDocumentGrounding`, i.e. an uploaded reference file (line ~1005).

Net effect: custom context between 1,200 and 4,000 characters is in a dead zone — cut from the guaranteed pinned block, and unreachable through the retrieval path that is documented as covering it.

### Expected behavior

Custom Context that the UI accepted (up to 4,000 characters) should reach the model, or the UI should not accept more than the engine can deliver. Either way the two limits should agree, and the comment describing the retrieval fallback should reflect what retrieval actually does for custom-context-only users.

### Steps to Reproduce

1. Open Settings → Custom Context.
2. Paste roughly 4,000 characters of prose about a project — several paragraphs separated by blank lines, no imperative one-liners (so nothing is classified as `pinned` by `customContextClassifier.ts`).
3. Do not upload any Reference File.
4. Start a session and ask a natural question whose answer is in the second half of that text, e.g. `what did we work on this sprint?`.
5. The answer reports that it has no context for the question.
6. Optional confirmation: launch with `NATIVELY_RETRIEVAL_DIAGNOSTICS=1` and observe the chunks being scored below the 0.18 floor and dropped.

### Environment

- OS: macOS 26.5
- App Version: 2.8.4
- Node.js: v26.4.0
- npm: 11.17.0
- Electron: 43.1.0
- Providers tested: built-in provider and a custom OpenRouter provider (same behavior on both, which points at prompt assembly rather than the provider path)

### Possible fixes

Listing options rather than assuming which direction you'd prefer:

1. **Raise `PINNED_INSTRUCTIONS_MAX_CHARS` to match the `custom_context` layer budget** (600 tokens ≈ 2,400 chars). Smallest change, stays inside the allocation the architecture already reserves, and doubles the guaranteed portion. I've opened a PR for this one.
2. **Exempt `custom_context` sources from `MIN_RELEVANCE_SCORE`.** The blob is user-authored and UI-capped, so relevance filtering buys little — retrieval exists for large documents, not a 4,000-char note the user wrote specifically so the assistant would know it.
3. **Align the UI cap with whatever the engine guarantees**, so the character counter never promises more than gets delivered.
4. **Correct the comment** on `PINNED_INSTRUCTIONS_MAX_CHARS` either way, since the retrieval fallback it points to does not apply when no reference file is present.

Happy to extend the PR in whichever direction you prefer.

abhiyan100:
## Summary

Fixes #399

`ModesManager.PINNED_INSTRUCTIONS_MAX_CHARS` was 1,200 chars (~300 tokens), but `LAYER_BUDGET.custom_context` in `llm/contextRoute.ts` already reserves **600 tokens** (~2,400 chars) for this layer — so the always-pinned block was spending half of its own allocation.

The comment on the constant justified the cap by stating that anything longer "remains fully available to RETRIEVAL (reference-file path), so nothing is lost." That does not hold for a user who has Custom Context but no uploaded reference file: `customContext` only reaches retrieval through the same `MIN_RELEVANCE_SCORE = 0.18` floor as a reference file, and a conversational question against prose chunks scores below it. I ported the scoring path from `ModeContextRetriever` (`score = matches / sqrt(queryWords.size × uniqueChunkWords)`) and ran a 4,000-char prose context against `what did we work on this sprint?`:

| Scenario | queryWords.size | threshold | chunk scores | passed |
|---|---|---|---|---|
| No transcript | 5 | 0.1800 | 0.1227, 0.1729 | 0 of 2 |
| Live transcript | 100 | 0.1800 | 0.1098, 0.1546 | 0 of 2 |

Retrieval then returns `formattedContext: ''`, so text past the cap was unreachable rather than deferred. Full analysis in #399.

This PR takes the smallest fix that stays inside the existing budget design:

- `PINNED_INSTRUCTIONS_MAX_CHARS`: `1_200` → `2_400`, matching the 600-token allocation the layer already has.
- Corrects the misleading comment so it describes what retrieval actually does for custom-context-only users.
- Updates the stale `(1200)` reference in `ModeGenerator.ts`'s header comment.
- Updates the existing cap test and adds a regression test.

I deliberately did **not** raise the cap to the UI's 4,000-char limit, since that would exceed the reserved layer budget and is a product decision rather than a defect fix. #399 lists that broader option (raise the layer budget / align the UI counter / exempt `custom_context` from the relevance floor) for you to decide — happy to extend this PR in whichever direction you prefer.

## Type of Change
- 🐛 Bug Fix

## Testing & Environment
- [x] Manual test performed on: **macOS 26.5, app 2.8.4, Node v26.4.0, npm 11.17.0, Electron 43.1.0**

Verification:

- The new regression test in `ModePinnedInstructions.test.mjs` is sized to fail on the old constant and pass on the new one: its fixture is 1,379 chars, so the asserted tail (`lead with verification and benchmarking`) falls past the old 1,200-char cut and inside the new 2,400 one. Both bounds are asserted in the test itself, so the fixture can't silently drift out of that window.
- The fixture is multi-paragraph prose with no imperative openers and no sensitive terms, so it classifies entirely as `searchable` in `customContextClassifier` and exercises the non-`pinned` path that regressed.
- The existing cap test still asserts truncation, now against the 2,400 bound.
- Reproduced the original failure by asking a question whose answer sat past 1,200 chars of Custom Context with no reference file uploaded, and confirmed the dropped chunks with `NATIVELY_RETRIEVAL_DIAGNOSTICS=1`.

One note for reviewers: the test suite imports from `dist-electron/`, and I wasn't set up to run a full Electron/Rust build locally, so I verified the changed logic and the fixture bounds directly rather than through a built `npm test` run. Worth a CI run before merge.

### DIFF
--- a/electron/services/ModeGenerator.ts
+++ b/electron/services/ModeGenerator.ts
@@ -19,7 +19,7 @@
 //   the same local backend or a deterministic stub for plumbing checks.
 // - The generated `customContext` is what becomes the mode's "Real-time prompt"
 //   (ModesManager.getActiveModePinnedInstructions). It is capped at
-//   PINNED_INSTRUCTIONS_MAX_CHARS (1200) at INJECTION time, so we target well
+//   PINNED_INSTRUCTIONS_MAX_CHARS (2400) at INJECTION time, so we target well
 //   under that here and hard-trim as a safety net.
 // - Document grounding: a custom mode only becomes documentGroundedCustomModeActive
 //   when its customContext matches BOTH DOCUMENT_SOURCE_RE and
--- a/electron/services/ModesManager.ts
+++ b/electron/services/ModesManager.ts
@@ -1163,10 +1163,17 @@ export class ModesManager {
     }
 
     // Hard cap for the always-pinned "Real-time prompt" (mode customContext).
-    // Roughly 300 tokens — enough for real mode instructions, small enough that
-    // a pasted document can't crowd out the transcript. Anything longer remains
-    // fully available to RETRIEVAL (reference-file path), so nothing is lost.
-    private static readonly PINNED_INSTRUCTIONS_MAX_CHARS = 1_200;
+    // Sized to the budget this layer is already allocated: LAYER_BUDGET
+    // .custom_context in llm/contextRoute.ts reserves 600 tokens, which is
+    // ~2,400 chars, so pinning up to that spends the allocation without
+    // crowding the transcript. The previous 1,200 was ~300 tokens — half the
+    // reserved budget — and the remainder was NOT recoverable in practice: the
+    // comment here used to claim retrieval covered it, but customContext only
+    // reaches retrieval through the same MIN_RELEVANCE_SCORE floor as reference
+    // files, and a conversational question against prose chunks scores below
+    // that floor (see ModeContextRetriever's adaptive-threshold notes), so a
+    // user with no uploaded reference file lost everything past the cap.
+    private static readonly PINNED_INSTRUCTIONS_MAX_CHARS = 2_400;
 
     /**
      * PI v3 (W2): the active mode's user-authored "Real-time prompt"
--- a/electron/services/__tests__/ModePinnedInstructions.test.mjs
+++ b/electron/services/__tests__/ModePinnedInstructions.test.mjs
@@ -70,13 +70,46 @@ describe('W2: getActiveModePinnedInstructions', () => {
         assert.match(nego, /180k/);
     });
 
-    test('caps at ~1,200 chars', () => {
+    test('caps at ~2,400 chars (the custom_context layer budget)', () => {
         const mgr = installActiveMode(makeMode({
             customContext: 'pitch the integration story. '.repeat(200),
         }));
         const pinned = mgr.getActiveModePinnedInstructions('sales_answer');
-        assert.ok(pinned.length <= 1_300, `len=${pinned.length}`);
+        assert.ok(pinned.length <= 2_500, `len=${pinned.length}`);
         assert.match(pinned, /\[truncated\]/);
+    });
+
+    // Regression: a multi-paragraph prose context under the cap must survive
+    // WHOLE. Everything past the old 1,200-char cap was unreachable — it is
+    // dropped from this block, and retrieval cannot rescue it because
+    // customContext is scored against MIN_RELEVANCE_SCORE like a reference
+    // file, which a conversational question does not clear.
+    test('prose context within the cap is pinned in full, including its tail', () => {
+        // The fixture is deliberately sized so its LAST paragraph sits between
+        // the old 1,200-char cap and the current 2,400 one: this test fails on
+        // the old constant and passes on the new one.
+        const filler = paragraph =>
+            `${paragraph} The platform answers impact questions about public data products, gathers evidence, and produces a cited report for each request it receives.`;
+        const paragraphs = [
+            filler('OVERVIEW.'),
+            filler('VERIFICATION: every claim is checked against the source it cites before publication.'),
+            filler('BENCHMARK: reports are graded against curated anchor facts per product.'),
+            filler('RELIABILITY: a failed run is retried automatically instead of ending the session.'),
+            filler('REFRESH: the source catalogue is rebuilt by a locked job so it cannot double-run.'),
+            filler('SCORING: the impact score is being redefined so it no longer saturates early.'),
+            'NARRATIVE: when asked what changed this cycle, lead with verification and benchmarking rather than new surfaces.',
+        ].join('\n\n');
+
+        assert.ok(paragraphs.length > 1_200, `fixture must exceed the old cap: ${paragraphs.length}`);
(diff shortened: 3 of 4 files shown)


abhiyan100:
Thanks for the review. I dug into this one and I don't think the stated mechanism holds — but it pointed at something worth documenting, so I've pushed a follow-up commit.

**`block.tokenBudget` is never enforced.** `enforceTokenBudget()` (PromptAssembler.ts:501) trims against the packet's *global* `params.tokenBudget`, walking blocks in `TRUST_LEVEL_ORDER`. It never reads the per-block field:

```ts
for (const block of sortedBlocks) {
    const blockTokens = this.estimateTokens(block.content);   // measured, not block.tokenBudget
    if (totalTokens + blockTokens > maxTokens && keptBlocks.length > 0) { … }
```

`addBlock()` (line 367) is a bare `push` with no truncation, and a repo-wide search for `block.tokenBudget` returns 0 reads — the field is set in ~14 places and consumed nowhere. So the `300` on this block was not capping the pinned content; the accessor's char cap was the only limit in play.

**The crowding concern is also already handled**, which I think is the more interesting half. `WhatToAnswerLLM.ts:483` builds the budget as:

```ts
const assemblerBudget = 2000
    + estimateTokens(intentContext || '')
    + estimateTokens(modeContextBlock)
    + estimateTokens(pinnedModeInstructions)   // ← grows with this block
    + …
```

The pinned block is self-funding: the global budget expands by exactly what the block adds, and `reservedForFit` then fits the transcript around it. So going from ~300 to ~600 tokens of pinned context does not displace the transcript, even though `MODE_POLICY` outranks `UNTRUSTED_TRANSCRIPT` in the trim order. That was the original cap's stated worry ("small enough that a pasted document can't crowd out the transcript") and it's structurally covered here.

**What I did change:** the advisory figure was stale in a way that would bite whoever reads it next, or anyone who later makes the field authoritative — at `300` it described half of what the block now carries. Pushed in 711b7d7: set it to `600` to match both the accessor's 2,400-char cap and `LAYER_BUDGET.custom_context`, with a comment noting the field is advisory so the next reader doesn't assume it's enforced.

If you'd rather `block.tokenBudget` became authoritative in `enforceTokenBudget()`, that's a larger change than this fix — several blocks declare figures well above what they'd get under a global trim (1200/3000/4000), so making it binding would change assembly behavior repo-wide. Happy to open that separately if it's a direction you want.

Issue #331, opened 2026-06-21
[Bug]: Whisper Large V3 install fails on macos


psiberfunk:
### What happened?

Fresh Natively install on an M3 mac and I get this error message on trying to install Whisper V3 large via the UX:

Failed to load model: Exception during initialization: filesystem error: in file_size: No such file or directory ["/Users/xxxx/Library/Application Support/natively/whisper-models/onnx-community/whisper-large-v3-turbo-ONNX/onnx/encoder_model.onnx_data"]

<img width="867" height="297" alt="Image" src="https://github.com/user-attachments/assets/3186c894-5fe8-4f28-9623-3c7497b2c869" />

### Steps to Reproduce

Open Natively, go to settings, go to Audio, Select Local Whisper as the mode, try to install the Whisper Large V3 Turbo model, see error.

### Environment

MacOS 26.5.1 , Natively Version 2.7.0 (2.7.0)

### Relevant Log Output

```shell
Failed to load model: Exception during initialization: filesystem error: in file_size: No such file or directory ["/Users/xxxx/Library/Application Support/natively/whisper-models/onnx-community/whisper-large-v3-turbo-ONNX/onnx/encoder_model.onnx_data"]
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

psiberfunk:
Note: Other models install fine, this does not appear to be a disk space error nor a file system error, as the cited directory exists and is clearly writable and has the current contents after my attempt to install:

<img width="674" height="294" alt="Image" src="https://github.com/user-attachments/assets/c8106498-1ae4-4dc1-a029-4058dd4eca33" />

I tried quitting Natively, destroying this folder and re-installing.. and it refused to work.  Annoyingly, LEAVING the folder/files there tricks Natively into thinking it IS installed when clearly it hasn't been downloaded.. so it's a bit of an evil/misleading bug if you don't catch it right away.

bhargavap21:
## Resolution

This was fixed in `aac3162` / `2398cb9` (`fix(whisper): download external-data weights for split ONNX checkpoints`) and shipped in **v2.8.1+**.

[...]

PR #403, opened 2026-07-28
fix(whisper): never load an fp32 encoder above PartitionAlloc's 2 GiB ceiling

### LINKED ISSUE #402: [Bug]: Whisper Large v3 Turbo hard-crashes the app (SIGTRAP) — 2 GiB fp32-encoder allocation exceeds Chromium PartitionAlloc's hard cap
### What happened?

Loading **Whisper Large v3 Turbo** hard-crashes the whole app on Apple Silicon, 100% reproducibly, regardless of how much RAM is free. Not a caught error, not an "STT failed" toast — the process dies with `EXC_BREAKPOINT (SIGTRAP)` a few seconds after launch, so transcription never works and there is no in-app diagnostic. Because the model preloads at startup, the app becomes unusable: it dies before you can reach Settings to pick a different model.

The model is fully downloaded and intact (this is *not* #331 — the `encoder_model.onnx_data` companion is present):

```
whisper-models/onnx-community/whisper-large-v3-turbo-ONNX/onnx/
  encoder_model.onnx          429K
  encoder_model.onnx_data     2.4G   ← fp32 encoder weights
  decoder_model_merged.onnx   656M
```

The crashing thread is the ONNX Runtime session init, and it dies inside the allocator:

```
InferenceSessionWrap::LoadModel
OrtApis::CreateSession
onnxruntime::InferenceSession::Initialize
onnxruntime::session_state_utils::SaveInitializedTensors
onnxruntime::BFCArena::AllocateRawInternal → BFCArena::Extend
onnxruntime::CPUAllocator::Alloc
_posix_memalign → _malloc_zone_memalign          ← allocation fails
Electron Framework (OOM handler) → brk 0         ← SIGTRAP
```

**This is not an out-of-memory condition.** The requested block size is in the register state: `x19 = x21 = x24 = 0x80000000` = **exactly 2 GiB in one allocation**. Chromium's PartitionAlloc caps a single allocation at

```cpp
// partition_alloc_constants.h
MaxDirectMapped() = (1UL << 31) - kSuperPageSize   // 2 GiB − 2 MiB
// "Intentionally set to less than 2GiB to make sure that a 2GiB allocation
//  fails. This is a security choice in Chrome, to help making size_t vs int
//  bugs harder to exploit."
```

A 2 GiB request is therefore **guaranteed** to fail in any Electron process, on any machine, at any memory pressure — and PartitionAlloc does not return `NULL`, it intentionally traps. Hence `brk 0` / SIGTRAP with no catchable JS error.

Evidence it is the cap and not memory:

- Crashes identically with 5.3 GB free + 2.9 GB reclaimable (82% system-wide free) and with almost nothing free. Closing every other app changes nothing.
- The **same** `encoder_model.onnx` + `encoder_model.onnx_data`, with the **same** `onnxruntime-node@1.22.0`, loads in **1.3–1.8 s** under plain Node 23 (system malloc) on the same machine — both with `enableCpuMemArena: true` and `false`. Only inside Electron does it die.

**Root cause chain in this repo:**

1. `electron/audio/whisper/inferenceConfig.ts` — `WHISPER_SAFE_DTYPE` pins `encoder_model: 'fp32'` for every Whisper-family model (deliberate, to protect encoder WER). For Large v3 Turbo that is a 2.4 GB fp32 encoder.
2. ORT's BFC arena grows by doubling, so `BFCArena::Extend` asks for a single 2 GiB chunk to hold those initializers — landing exactly on PartitionAlloc's hard ceiling.
3. Nothing guards this: `electron/audio/whisper/modelManager.ts` has no memory or allocation-size gating (`requiresAppleSilicon` is the only hardware gate, no `os.totalmem()` anywhere), and there is no way to opt the *encoder* down to q8/q4 — `whisperAppleSiliconDtype` only goes the other direction, back to uniform fp32.

**Suggested fixes** (in rough order of value):

- **Disable the CPU mem arena for the whisper session** (`enableCpuMemArena: false` in the ORT session options). Without the arena, ORT allocates per-tensor instead of one doubling chunk, so no single request approaches 2 GiB. Verified locally: the same encoder loads fine that way (1.3 s), and it is a one-line change.
- Cap or chunk the arena so no single allocation can reach `2 GiB − 2 MiB`, and pre-flight the load: if the resolved dtype implies a larger single allocation, fail with a real `WorkerErrorResponse` before calling `CreateSession`. Anything that reaches `posix_memalign` with ≥2 GiB is unrecoverable by construction.
- Offer a quantized encoder (q8/q4) for the large checkpoints even if fp32 stays default for the small ones — a q8 encoder that loads beats an fp32 encoder that traps.
- Don't preload the selected model on startup without a fallback: right now one bad model choice makes the app crash-loop with no way in to change it. (My only workaround was editing `localWhisperModel` in `settings.json` by hand.)
- Secondary: the catalog lists `sizeMb: 1031` for this model, but on-disk with the fp32 encoder it is **3.0 GB** — off by ~3×.

### Steps to Reproduce

1. On an Apple Silicon Mac, install Natively 2.7.0.
2. Settings → Audio → Local Whisper.
3. Install / select **Whisper Large v3 Turbo** and let the download finish (verify `encoder_model.onnx_data` is ~2.4 GB, so this isn't #331).
4. Start transcription (or just let the model preload).
5. App terminates with `EXC_BREAKPOINT (SIGTRAP)` a few seconds after launch. macOS crash reporter shows the ONNX Runtime `CreateSession` stack above. No in-app error, nothing in `logs/telemetry.jsonl`.
6. Relaunch → same crash, every time. Free memory makes no difference. Since the model preloads on startup, you cannot reach Settings to switch models.

Smaller models (base / small / medium) load and transcribe fine on the same machine.

### Environment

- Natively 2.7.0 (2.7.0), `com.electron.meeting-notes`, ARM-64 native
- Electron Framework 33.4.11
- macOS 26.3 (25D125)
- Mac14,9 (MacBook Pro 14", M2 Pro), 16 GB RAM (reproduced both under pressure and with 82% of RAM free)
- onnxruntime 1.22.0 (`libonnxruntime.1.22.0.dylib`)
- Model: `onnx-community/whisper-large-v3-turbo-ONNX`, fully downloaded (3.0 GB on disk)

### Relevant Log Output

```shell
Process:             Natively [14408]
Identifier:          com.electron.meeting-notes
Version:             2.7.0 (2.7.0)
OS Version:          macOS 26.3 (25D125)
Hardware Model:      Mac14,9

Launch Time:         2026-07-28 13:44:35.7534 +0700
Date/Time:           2026-07-28 13:44:49.3923 +0700

Exception Type:      EXC_BREAKPOINT (SIGTRAP)
Exception Codes:     0x0000000000000001, 0x000000010f39db60
Termination Reason:  Namespace SIGNAL, Code 5, Trace/BPT trap: 5
Triggered by Thread: 37

Thread 37 Crashed:
0   Electron Framework            0x10f39db60 <OOM handler / trap>
...
8   libsystem_malloc.dylib        0x185c30654 _malloc_zone_memalign + 320
9   libsystem_malloc.dylib        0x185c30d3c _posix_memalign + 52
10  libonnxruntime.1.22.0.dylib   0x14c3c778c onnxruntime::CPUAllocator::Alloc(unsigned long) + 60
11  libonnxruntime.1.22.0.dylib   0x14c3c8dd0 onnxruntime::BFCArena::Extend(unsigned long) + 368
12  libonnxruntime.1.22.0.dylib   0x14c3c9b98 onnxruntime::BFCArena::AllocateRawInternal(...) + 916
13  libonnxruntime.1.22.0.dylib   0x14c3c976c onnxruntime::BFCArena::Alloc(unsigned long) + 60
14  libonnxruntime.1.22.0.dylib   0x14c464a18 onnxruntime::Tensor::Tensor(...) + 136
15  libonnxruntime.1.22.0.dylib   0x14c44dec0 onnxruntime::session_state_utils::AllocateTensorOnDeviceOrMemory(...) + 236
16  libonnxruntime.1.22.0.dylib   0x14c44f85c onnxruntime::session_state_utils::SaveInitializedTensors(...) + 5696
17  libonnxruntime.1.22.0.dylib   0x14c4433f4 onnxruntime::SessionState::FinalizeSessionStateImpl(...) + 1556
20  libonnxruntime.1.22.0.dylib   0x14ba83cbc onnxruntime::InferenceSession::Initialize() + 5896
21  libonnxruntime.1.22.0.dylib   0x14bafbce4 InitializeSession(...) + 2004
22  libonnxruntime.1.22.0.dylib   0x14bac5764 OrtApis::CreateSession(...) + 104
23  onnxruntime_binding.node      0x1499a3038 InferenceSessionWrap::LoadModel(Napi::CallbackInfo const&) + 808
...
45  libsystem_pthread.dylib       0x185e03c08 _pthread_start + 136

Thread 37 crashed with ARM Thread State (64-bit):
   x19: 0x0000000080000000   x21: 0x0000000080000000   x24: 0x0000000080000000   ← 2 GiB requested
   esr: 0xf2000000 (Breakpoint) brk 0
```

`logs/telemetry.jsonl` contains no record of the failure — the process dies before anything is logged.

Tarasusrus:
## Symptom

Selecting **Whisper Large v3 Turbo** hard-crashes the whole app with `EXC_BREAKPOINT (SIGTRAP)` a few seconds after launch — not a caught error, not an "STT failed" toast. The model is fully downloaded and intact (this is not #331: `encoder_model.onnx_data` is present at 2.4 GB). Because the selected model preloads at startup, the app crash-loops and there is no way in to Settings to pick a different model; the only workaround is hand-editing `localWhisperModel` in `settings.json`. Nothing is written to `logs/telemetry.jsonl` — the process dies first.

## Root cause

`WHISPER_SAFE_DTYPE` in `electron/audio/whisper/inferenceConfig.ts` pins `encoder_model: 'fp32'` for every Whisper-family model. Turbo's fp32 encoder weights are ~2.4 GB, ONNX Runtime materialises those initializers in **one block**, and the request lands on Chromium's PartitionAlloc hard cap:

```cpp
// partition_alloc_constants.h
MaxDirectMapped() = (1UL << 31) - kSuperPageSize   // 2 GiB - 2 MiB
// "Intentionally set to less than 2GiB to make sure that a 2GiB allocation
//  fails. This is a security choice in Chrome, to help making size_t vs int
//  bugs harder to exploit."
```

PartitionAlloc does not return `NULL` at that point — it intentionally traps. Hence `brk 0` / SIGTRAP with no catchable JS error. The register state in the crash report is the smoking gun: `x19 = x21 = x24 = 0x80000000`, i.e. **exactly 2 GiB in one allocation**, requested from `BFCArena::Extend → CPUAllocator::Alloc → posix_memalign`.

This is **not** an out-of-memory condition:

- reproduces identically with 82% of system RAM free, and with almost nothing free;
- the same files with the same `onnxruntime-node@1.22.0` load in 1.3–1.8 s under plain Node on the same machine (system malloc). Only inside Electron does it die.

### Why `enableCpuMemArena: false` isn't the fix

The issue's first suggestion was to disable the CPU mem arena — but `electron/utils/onnxThreadConfig.ts` **already defaults `enableCpuMemArena: false`**, and the crash happens anyway. Turning the arena off removes the doubling-chunk behaviour, but it does not help when a *single fp32 initializer tensor* is itself at or above the ceiling: that one tensor still has to be allocated in one piece. Any allocation path that reaches `posix_memalign` with >= 2 GiB is unrecoverable by construction, so the only durable fix is to never make the request.

## What changed

Two layers, so a future caller can't reintroduce the trap:

1. **`electron/audio/whisper/allocationGuard.ts`** (new, pure module — no electron/fs/transformers, importable from the worker, the main process and unit tests):
   - `PARTITION_ALLOC_MAX_BYTES` (2 GiB − 2 MiB) and `exceedsPartitionAllocCap()`;
   - `guardEncoderDtype(dtype, encoderFp32Bytes)` rewrites **only** the `encoder_model` entry to `q8`, and only when the resolved encoder dtype is fp32 *and* the catalog records fp32 encoder weights at or above the cap. Decoder dtypes, execution providers, and models with no recorded size are returned untouched, so nothing that loads today can regress. A uniform `'fp32'` string is widened into a per-module map rather than quantizing the whole graph — silently dropping decoder precision as a side effect of an allocation fix would be an accuracy change nobody asked for.
   - `preflightAllocation(modelId, dtype, encoderFp32Bytes)` returns a human-readable refusal message, or `null` when the load is safe.
2. **`whisperWorker.ts`** calls `preflightAllocation` immediately before `pipeline()`; a non-null result is posted as a `WorkerErrorResponse` and the worker returns, so `CreateSession` is never reached. This is defence in depth for any caller that hand-builds an init message instead of using `buildWorkerInitMessage`.

Supporting changes:

- `modelManager.ts` records `encoderFp32Mb: 2400` for `onnx-community/whisper-large-v3-turbo-ONNX` and exports `getModelEncoderFp32Bytes()`. Only checkpoints where the size actually matters carry the field; `undefined` means "unknown" and is treated as safe.
- `inferenceConfig.ts` `buildWorkerInitMessage()` applies the guard after `resolveInferenceConfig()`, logs one `console.warn` with the reason when it downgrades, and forwards both the guarded dtype and `encoderFp32Bytes` in the init message. The lookup uses the same best-effort lazy-require + `try/catch` style as the existing size / external-data lookups, so a failure there can never block worker startup.
- `getAvailableModels()` now computes cache status against the dtype a model will *actually* load, so turbo is no longer reported "available" on the strength of an fp32 encoder it no longer uses.
- `types.ts` gains `WhisperModelInfo.encoderFp32Mb` and `WorkerInitMessage.encoderFp32Bytes`, both optional and documented.

Result on the reported machine's configuration: turbo resolves `encoder_model: 'q8'` with the decoders unchanged, so the 2 GiB request is never made. A q8 encoder that loads beats an fp32 encoder that traps.

Not addressed here (separate concerns from the crash): the catalog's `sizeMb: 1031` under-reporting the 3.0 GB on-disk footprint, and the lack of a startup fallback when the preloaded model fails.

## How it was tested

`electron/audio/__tests__/WhisperAllocationCeiling.test.mjs` (new, 10 tests, same conventions as the neighbouring `WhisperExternalDataFormat.test.mjs`): the cap constant, `exceedsPartitionAllocCap` boundaries (just under / at / over the cap, plus `0`/`undefined`/`NaN`/negative), encoder-only downgrade for turbo-sized weights with decoder entries proven untouched, no downgrade for small models / already-q8 encoders / unknown sizes, string→map widening, `preflightAllocation` refusing the unsafe case and passing the guarded one, `buildWorkerInitMessage` returning a non-fp32 encoder for turbo and an unchanged fp32 encoder for `Xenova/whisper-tiny.en`, and cache-status alignment.

```
npm run build:electron
node --test electron/audio/__tests__/WhisperAllocationCeiling.test.mjs   # 10/10 pass
npm run typecheck:electron                                              # clean
node --test 'electron/audio/__tests__/**/*.test.mjs'                    # 285/295 pass
node --test 'electron/audio/whisper/__tests__/**/*.test.mjs'            # 10/11 pass
```

The failures in the two suite-wide runs are pre-existing on `main` at `f25d2b8` — the identical 10 (`electron/audio/__tests__`) and 1 (`electron/audio/whisper/__tests__`) fail with this branch stashed, and none of them touch whisper inference config. This branch adds 10 passing tests and changes no existing result.

**Mutation check.** Making `guardEncoderDtype` return its input unchanged turns 5 of the 10 new tests RED (`guardEncoderDtype downgrades ONLY the encoder…`, `…widens a uniform fp32 string…`, `preflightAllocation refuses…`, `buildWorkerInitMessage never sends an fp32 encoder…`, `model status is computed for the dtype turbo will actually load`); restoring it returns 10/10 GREEN. Separately, making `preflightAllocation` always return `null` turns the worker-preflight test RED on its own, so the second layer is independently covered rather than riding on the first.

Fixes #402

### DIFF
--- a/electron/audio/__tests__/WhisperAllocationCeiling.test.mjs
+++ b/electron/audio/__tests__/WhisperAllocationCeiling.test.mjs
@@ -1,0 +1,219 @@
+// Regression test for the Whisper Large v3 Turbo SIGTRAP (issue #402, 2026-07-28).
+//
+// Chromium's PartitionAlloc refuses any SINGLE allocation of 2 GiB - 2 MiB or
+// more, and it does not return NULL — it traps:
+//
+//   MaxDirectMapped() = (1UL << 31) - kSuperPageSize   // partition_alloc_constants.h
+//
+// WHISPER_SAFE_DTYPE pins `encoder_model: 'fp32'` for every Whisper-family
+// model. Whisper Large v3 Turbo's fp32 encoder weights are ~2.4 GB
+// (encoder_model.onnx_data), ONNX Runtime materialises those initializers in
+// one block, and the request lands exactly on the ceiling: posix_memalign →
+// PartitionAlloc → `brk 0` → EXC_BREAKPOINT (SIGTRAP). The whole app dies with
+// no catchable JS error and nothing in telemetry, and because the selected
+// model preloads at startup the app crash-loops with no way into Settings.
+//
+// This is NOT out-of-memory (reproduces with 82% of RAM free) and it is NOT
+// fixed by arena settings — onnxThreadConfig already defaults
+// enableCpuMemArena:false and the crash happens anyway, because a single fp32
+// initializer tensor is itself over the cap.
+//
+// Guards the two-layer fix:
+//   1. buildWorkerInitMessage resolves the ENCODER (only) to q8 for checkpoints
+//      whose catalog fp32 encoder size is at or above the cap.
+//   2. preflightAllocation refuses such a load with a structured error before
+//      pipeline()/CreateSession, for callers that bypass layer 1.
+//
+// Runs against the esbuild/tsc-compiled modules in dist-electron/.
+// Run via: npm run build:electron && node --test electron/audio/__tests__/
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import path from 'node:path';
+import fs from 'node:fs';
+import os from 'node:os';
+import Module from 'node:module';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// modelManager + inferenceConfig both pull in `electron` via getModelsDir().
+const userData = fs.mkdtempSync(path.join(os.tmpdir(), 'whisper-alloc-'));
+const origLoad = Module._load;
+Module._load = function patched(request, _p, _m) {
+  if (request === 'electron') {
+    return { app: { getPath: () => userData, isReady: () => true } };
+  }
+  return origLoad.apply(this, arguments);
+};
+
+const guardPath = path.resolve(
+  __dirname,
+  '../../../dist-electron/electron/audio/whisper/allocationGuard.js',
+);
+const modelMgrPath = path.resolve(
+  __dirname,
+  '../../../dist-electron/electron/audio/whisper/modelManager.js',
+);
+const inferenceCfgPath = path.resolve(
+  __dirname,
+  '../../../dist-electron/electron/audio/whisper/inferenceConfig.js',
+);
+
+const {
+  PARTITION_ALLOC_MAX_BYTES,
+  ENCODER_FALLBACK_DTYPE,
+  exceedsPartitionAllocCap,
+  guardEncoderDtype,
+  preflightAllocation,
+} = await import(pathToFileURL(guardPath).href);
+const { getModelEncoderFp32Bytes, getAvailableModels } = await import(
+  pathToFileURL(modelMgrPath).href
+);
+const { buildWorkerInitMessage } = await import(pathToFileURL(inferenceCfgPath).href);
+
+const TURBO = 'onnx-community/whisper-large-v3-turbo-ONNX';
+const TINY_EN = 'Xenova/whisper-tiny.en';
+
+// The per-module map the resolver produces on every platform.
+const SAFE_DTYPE = {
+  encoder_model: 'fp32',
+  decoder_model: 'q8',
+  decoder_model_merged: 'q8',
+  decoder_with_past_model: 'q8',
+};
+
+const GIB = 1024 ** 3;
+
+test('the cap is Chromium PartitionAlloc MaxDirectMapped: 2 GiB - 2 MiB', () => {
+  assert.equal(PARTITION_ALLOC_MAX_BYTES, 2 ** 31 - 2 * 1024 * 1024);
+  // Sanity: the crashing request in the report was exactly 0x80000000 (2 GiB),
+  // which must be over the cap.
+  assert.equal(exceedsPartitionAllocCap(0x80000000), true);
+});
+
+test('exceedsPartitionAllocCap: boundary is inclusive, unknown sizes are safe', () => {
+  assert.equal(exceedsPartitionAllocCap(PARTITION_ALLOC_MAX_BYTES - 1), false);
+  assert.equal(exceedsPartitionAllocCap(PARTITION_ALLOC_MAX_BYTES), true);
+  assert.equal(exceedsPartitionAllocCap(PARTITION_ALLOC_MAX_BYTES + 1), true);
+  // Unknown / absent / nonsense sizes must never block a load that works today.
+  assert.equal(exceedsPartitionAllocCap(0), false);
+  assert.equal(exceedsPartitionAllocCap(undefined), false);
+  assert.equal(exceedsPartitionAllocCap(null), false);
+  assert.equal(exceedsPartitionAllocCap(NaN), false);
+  assert.equal(exceedsPartitionAllocCap(-1), false);
+});
+
+test('guardEncoderDtype downgrades ONLY the encoder for turbo-sized fp32 weights', () => {
+  const res = guardEncoderDtype(SAFE_DTYPE, Math.round(2400 * 1024 * 1024));
+  assert.equal(res.downgraded, true);
+  assert.equal(res.dtype.encoder_model, ENCODER_FALLBACK_DTYPE);
+  assert.equal(ENCODER_FALLBACK_DTYPE, 'q8');
+  // Decoder entries are untouched — this fix must not silently change decoder
+  // precision, and it must not drop keys the loader relies on.
+  assert.equal(res.dtype.decoder_model, 'q8');
+  assert.equal(res.dtype.decoder_model_merged, 'q8');
+  assert.equal(res.dtype.decoder_with_past_model, 'q8');
+  assert.ok(res.reason && res.reason.length > 0);
+  // Input must not be mutated in place.
+  assert.equal(SAFE_DTYPE.encoder_model, 'fp32');
+});
+
+test('guardEncoderDtype leaves small models, already-quantized encoders and unknown sizes alone', () => {
+  // Small model (tiny.en fp32 encoder is a few tens of MB).
+  const small = guardEncoderDtype(SAFE_DTYPE, 40 * 1024 * 1024);
+  assert.equal(small.downgraded, false);
+  assert.deepEqual(small.dtype, SAFE_DTYPE);
+
+  // Encoder already quantized → nothing to downgrade, even at turbo size.
+  const q8Encoder = { ...SAFE_DTYPE, encoder_model: 'q8' };
+  const already = guardEncoderDtype(q8Encoder, 3 * GIB);
+  assert.equal(already.downgraded, false);
+  assert.deepEqual(already.dtype, q8Encoder);
+  const uniformQ8 = guardEncoderDtype('q8', 3 * GIB);
+  assert.equal(uniformQ8.downgraded, false);
+  assert.equal(uniformQ8.dtype, 'q8');
+
+  // Size unknown → treated as safe (no catalog figure must never break a load).
+  const unknown = guardEncoderDtype(SAFE_DTYPE, undefined);
+  assert.equal(unknown.downgraded, false);
+  assert.deepEqual(unknown.dtype, SAFE_DTYPE);
+});
+
+test('guardEncoderDtype widens a uniform fp32 string into a map instead of quantizing everything', () => {
+  const res = guardEncoderDtype('fp32', 3 * GIB);
+  assert.equal(res.downgraded, true);
+  assert.equal(typeof res.dtype, 'object');
+  assert.equal(res.dtype.encoder_model, ENCODER_FALLBACK_DTYPE);
+  // The rest of the graph keeps the requested fp32 — quantizing the decoder as
+  // a side effect of an allocation fix would be a silent accuracy change.
+  assert.equal(res.dtype.decoder_model, 'fp32');
+  assert.equal(res.dtype.decoder_model_merged, 'fp32');
+  assert.equal(res.dtype.decoder_with_past_model, 'fp32');
+});
+
+test('preflightAllocation refuses the untenable load and passes everything else', () => {
+  // Layer-2: an init message hand-built with an fp32 encoder for turbo.
+  const msg = preflightAllocation(TURBO, SAFE_DTYPE, Math.round(2400 * 1024 * 1024));
+  assert.equal(typeof msg, 'string');
+  assert.match(msg, /whisper-large-v3-turbo/);
+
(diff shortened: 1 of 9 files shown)


Tarasusrus:
Good catch — valid, and worse than "inconsistent": it was permanent. Fixed in 2e60256.

The guard was only applied in `buildWorkerInitMessage` and `getAvailableModels`. Every other consumer that maps a model to on-disk artifacts read the bare platform dtype from `resolveInferenceConfig()`, so they looked for `encoder_model.onnx` while the worker had actually downloaded `encoder_model_quantized.onnx`. For turbo that meant a *complete* download verified as incomplete — and re-downloading produced the same q8 files, so it never converged — while a stale pre-guard fp32 cache was reported ready and preloaded at startup.

Rather than patching each call site, the guarded per-model dtype is now the single source of truth: `resolveModelDtype(modelId)` is exported from `inferenceConfig.ts`, backed by the same `resolveGuardedDtype` helper that builds the worker init message. So the dtype the worker is sent and the dtype the cache is checked against cannot drift apart again. Routed through it:

- `electron/main.ts` — startup preload readiness (now per-model, since the guard is per-model)
- `electron/ipcHandlers.ts` — `local-whisper-preload`
- `electron/services/LocalModelDownloadService.ts` — whisper provider `isModelCached` (download verification)
- `modelManager.getAvailableModels` + `buildWorkerInitMessage`

Contracts preserved: every lookup stays best-effort (unknown id or failing catalog lookup degrades to the platform dtype, never blocks a load or download), `isModelCached(id)` with no dtype still falls through to the legacy directory check, and the `inferenceConfig` <-> `modelManager` requires stay lazy on both sides so no import cycle is introduced.

Verification:
- `npm run typecheck:electron` and `npm run build:electron` clean.
- `WhisperAllocationCeiling.test.mjs` 10 -> 15 tests, all green. New coverage: a q8-encoder turbo cache is READY via both `isModelCached` and the download provider (your point about the service not being covered), a stale fp32 cache is NOT, `resolveModelDtype` diverges from the unguarded platform dtype for turbo only, and it always agrees with what `buildWorkerInitMessage` sends.
- Neighbouring whisper suites unchanged and green: WhisperExternalDataFormat 9/9, WhisperLoadSentinel 6/6, WhisperProgressAggregator 14/14, LocalWhisper* all green. No new failures anywhere against the pre-existing baseline.
- Mutation-checked: making `resolveModelDtype` return the unguarded dtype turns all four new tests red (plus the existing model-status one); restoring turns them green. The tests fail without the fix.

Tarasusrus:
Closing this — my premise was wrong and I want that on the record rather than left implied.

I re-tested in an isolated harness (Electron 33.4.11 + @huggingface/transformers 3.8.1 + onnxruntime-node 1.22.0, same on-disk checkpoint), loading the model in a worker thread exactly as whisperWorker does:

- No `session_options` (ORT's CPU arena left enabled): **SIGTRAP**, identical to the crash in #402. Reproduces on `executionProviders: ['coreml','cpu']` and on `['cpu']` alike, so CoreML is not involved.
- `session_options: { enableCpuMemArena: false }`: the session is created normally. The 2.4 GB fp32 encoder (`encoder_model.onnx` + `encoder_model.onnx_data`) loads in ~7.4 s, no trap.

So the trap is the BFC arena doubling its way to a >= 2 GiB request, not a single fp32 initializer that is untenable by construction. `getBoundedOnnxSessionOptions()` on `main` already defaults `enableCpuMemArena` to false, which means the crash this PR claims to fix does not occur on `main` — it occurs on the 2.7.0 release, whose worker passes no session options at all. The allocation guard here is therefore solving a problem I never demonstrated in this codebase, and I would rather withdraw it than land a fix justified by a stack trace from a build that predates the arena change.

Two things I did find that may be worth separate issues, if maintainers agree they are real:

1. Any ORT error raised during `CreateSession` inside the whisper worker aborts the process (`Napi::Error::ThrowAsJavaScriptException` → `__cxa_throw` → `abort`, SIGABRT) instead of surfacing as a catchable `WorkerErrorResponse`. On 2.7.0 this turns every model-load failure into a hard crash.
2. Whisper Large v3 Turbo does not run with transformers 3.8.1 regardless of dtype: inference fails with `Missing the following inputs: cache_position` on both the fp32 and the q8 merged decoder. `main` pins `^3.8.1`, so the model appears to be unusable there too, independent of anything in this PR.

Apologies for the noise on #402.

PR #418, opened 2026-08-03
fix(stealth): tray/taskbar follow stealth state, restore launcher pre-paint gate

evinjohnn:
Supersedes the stealth portion of #406. That PR is ~89 commits stale and does not apply to `main`; this branch is cut from current `main` and contains **only** the stealth fixes — none of its unrelated changes (STT low-RAM path, audio-device fallback rewrite, keybinds sub-tab UI, language/accent grid).

## What was broken

**1. Windows never hid the tray during Undetectable mode.** The only `hideTray()` call site lives inside `_enforceDockState()`, which begins `if (process.platform !== 'darwin') return;`. So on Windows there was effectively no call site at all — toggling Undetectable ON left a tray icon showing the real app name for the entire stealth session. This was a pre-existing gap on `main`, not something #406 introduced.

**2. macOS could skip the tray in both directions.** The `hideTray()` / `showTray()` calls sat inside the `shouldApply` guard, where `shouldApply = settled !== lastApplied` (`toggleStateReducer.ts`). Stealth ON while the dock was *already* hidden gave `shouldApply=false`, so `hideTray()` never ran — the same identity leak as (1). The mirror case (an OFF that skipped the dock transition) left no tray at all, and since `hideTray()` nulls `this.tray`, nothing re-minted it.

**3. The launcher lost its pre-paint gate.** `show: false` had been replaced with a hardcoded `skipTaskbar: false`. `show` defaults to `true`, so the window appeared before the renderer painted — reintroducing the white-screen flash that option exists to prevent — and its `ready-to-show` handler became dead code.

Verified deterministically rather than by eye, same build, one line changed:

| Build | `launcher.isVisible()` at construction |
|---|---|
| without `show: false` | `true` — on screen, pre-paint |
| with `show: false` | `false` — waits for `ready-to-show` |

**4. `setSkipTaskbar(false)` was unconditional**, so stealth never removed the launcher's taskbar entry — it stayed listed and Alt-Tab-able under its real title.

**5. The tray refresh pushed the dock icon into the tray.** `_applyDisguise()` resolves mode `none` to `natively.icns` on macOS; setting that as the tray image replaced the monochrome menu-bar template with a full-colour icon that no longer inverts for light/dark, and `showTray()`'s `if (this.tray) return` meant it stayed wrong until restart.

## What changed

- Tray follows the **desired** state, outside the dock predicate, on both platforms. Both calls are idempotent so enforcement retries stay safe.
- Added the missing win32 tray branch in `setUndetectable()`; reverted the startup path to `if (!getUndetectable()) showTray()` so a persisted-undetectable cold start never mints a tray.
- Restored `show: false`; `skipTaskbar` now tracks `getUndetectable()` instead of being a constructor constant, and `setUndetectable()` drives it from the toggle state.
- Extracted `buildTrayImage()` as the single source of tray icon resolution + template flag, shared by `showTray()` and `_applyDisguise()`.
- `getTrayTooltip()` derives from the disguise mode directly rather than reading `process.title` back (which silently depended on `_applyDisguise()` having run first).
- Dropped four `setSkipTaskbar` calls that were provable no-ops — the overlay and settings windows are constructed `skipTaskbar: true` and nothing sets them false.

## Validation

- `Tested physically on Windows` — cold start with persisted undetectable mints no tray; toggle ON destroys it; OFF re-creates it; launcher leaves the taskbar and Alt-Tab.
- `Covered by automated tests` — 47 existing stealth/dock/toggle tests pass, including `UndetectableDockLeakOnLauncherShow` (which exercises dropped-hide convergence, the mechanism touched here).
- `npx tsc --noEmit` and `tsc -p electron/tsconfig.json --noEmit` clean.
- **`Requires physical macOS verification`** — the macOS tray-ordering fix is reasoned from the dock predicate's semantics and covered only by the existing fakes. It has not been run on a Mac.

## Known gaps, deliberately not addressed here

- No re-assert path for Windows taskbar/tray state: `reassertUndetectableStealth()` returns early on non-darwin and covers only dock visibility and content protection. An `explorer.exe` restart or the AUMID re-registration in `setDisguise()` can restore the launcher's taskbar entry mid-session.
- Disguise mode → name/icon mappings still exist in three places (`buildTrayImage`, `_applyDisguise`, `WindowHelper`); consolidating them needs `DisguiseMode` exported out of `main.ts`.
- No new tests for the tray/taskbar branches. `getTrayTooltip()` reads `process.platform` internally, so neither branch is assertable without a signature change.

### DIFF
--- a/electron/WindowHelper.ts
+++ b/electron/WindowHelper.ts
@@ -417,6 +417,10 @@ export class WindowHelper {
         webSecurity: !isDev, // DEBUG: Disable web security only in dev
       },
       show: false, // DEBUG: Force show -> Fixed white screen, now relies on ready-to-show
+      // Taskbar presence must track stealth state, not be a hardcoded constant:
+      // a cold start with undetectable persisted must not paint a taskbar entry
+      // before setUndetectable() ever runs. Re-asserted on every toggle.
+      skipTaskbar: this.appState.getUndetectable(),
       // Platform-specific frame settings
       ...(isMac
         ? { titleBarStyle: 'hiddenInset' as const, trafficLightPosition: { x: 14, y: 14 } }
@@ -1096,6 +1100,7 @@ export class WindowHelper {
   public syncOverlayInteractionPolicy(quiet = false): void {
     if (!this.overlayWindow || this.overlayWindow.isDestroyed()) return;
 
+
     const passthrough = this.appState.getOverlayMousePassthrough();
     // The pill and toggle windows follow ONLY the stealth passthrough — they
     // are fully painted (no transparent margins), so the hover gate never
@@ -1125,7 +1130,6 @@ export class WindowHelper {
       this.overlayWindow.setIgnoreMouseEvents(true, { forward: true });
     } else {
       this.overlayWindow.setIgnoreMouseEvents(false);
-      // Restore full interactivity when capturing clicks.
       this.overlayWindow.setFocusable(true);
     }
     auxWindows.forEach((w) => {
@@ -1813,7 +1817,6 @@ export class WindowHelper {
   public showOverlay(): void {
     if (!this.overlayWindow || this.overlayWindow.isDestroyed()) return;
 
-    // Restore opacity in case it was zeroed by hideMainWindow() before a screenshot.
     this.overlayWindow.setOpacity(1);
     this.pillWindow?.setOpacity(1);
     this.toggleWindow?.setOpacity(1);
@@ -2007,6 +2010,7 @@ export class WindowHelper {
         expanded: true,
       });
 
+  
       // Restore opacity before showing (it may have been zeroed by hideMainWindow).
       if (process.platform === 'win32' && this.contentProtection) {
         // Opacity Shield: Show at 0 opacity first to prevent frame leak.
--- a/electron/WindowHelper.ts
+++ b/electron/WindowHelper.ts
@@ -2,6 +2,7 @@ import { app, BrowserWindow, Menu, screen } from 'electron';
 import fs from 'node:fs';
 import path from 'node:path';
 import { AppState } from './main';
+import { disguiseIconPath, appIconPath, normalizeDisguiseMode } from './disguise';
 import { KeybindManager } from './services/KeybindManager';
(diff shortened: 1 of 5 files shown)


evinjohnn:
Pushed `711ecdee`, closing the two gaps listed as known limitations on the first commit.

**Windows re-assert path.** `reassertUndetectableStealth()` returned early on non-darwin, so the launcher-show hook did nothing there — and the two things carrying the app's identity on Windows can come back on their own mid-session (taskbar button after an `explorer.exe` restart; the AUMID re-registration `setDisguise()` performs while stealth is on). Since the user has no reason to re-toggle, the entry would sit there under the real title for the rest of the session. Windows now re-asserts `skipTaskbar`, the hidden tray, and content protection. The darwin path is untouched — it still runs the self-verifying dock loop.

**Disguise identity consolidation.** The mode→name and mode→icon mappings existed in three copies, and the unpackaged branch had already drifted: one resolved via `path.resolve(__dirname, '../../assets/...')`, the others via `path.join(app.getAppPath(), ...)`.

Worth noting those *do* resolve to the same path today — but only because the esbuild output happens to sit exactly two levels below the repo root. It's latent fragility rather than a live bug: change the bundle layout and one copy breaks while the others don't. All three now come from `electron/disguise.ts`, anchored on `getAppPath()`, with lookups exhaustive over `DisguiseMode` so adding a mode is a compile error at each site instead of a silent fallthrough to the wrong icon.

**Tests.** `DisguiseIdentity.test.mjs` covers **both** platform branches — `platform` is an explicit parameter, so `darwin` and `win32` are asserted from a single run without mutating `process.platform` (the anti-pattern CLAUDE.md calls out). I checked the suite actually bites rather than just passing: mutating the win32 name mapping to return the macOS name fails 2 tests, and restoring it goes green.

59 stealth/dock/toggle/disguise tests pass; both typechecks clean.

Still `Requires physical macOS verification` — unchanged from the first commit. Everything here is Windows-verified only, and the macOS tray-ordering fix remains reasoned from the dock predicate's semantics plus existing fakes.

PR #426, opened 2026-08-03
fix(license): API-plan Pro survives an absent native module (premium bump + regression tests)

evinjohnn:
**Independent of the 7-PR stack.** Base: `main`.

> ### ⛔ Blocked on a submodule PR
> Merge **[natively-premium#2](https://github.com/Natively-AI-assistant/natively-premium/pull/2)** first.
> This PR bumps the `premium` pointer to `69da127`, which only exists on that branch today. Merging this first would leave the submodule reference dangling and break every fresh clone.

## Change summary

| File | Change |
|---|---|
| `premium` | Pointer `172f7add` → `69da127` |
| `electron/services/__tests__/LicenseNativeModuleAbsent.test.mjs` | New, 221 lines |
| `electron/services/__tests__/__electron_license_stub.mjs` | New, 38 lines |

The submodule bump carries two commits, both touching only `premium/electron/services/LicenseManager.ts`:

- `854ba57` — keep API-plan Pro working when the native module is absent
- `69da127` — do not protect a license proven to belong to another device

The pointer gap is otherwise empty: `git diff 172f7add origin/main` inside the submodule is zero bytes, so this bump carries those two commits' changes and nothing else.

## The bug being fixed

`storeLicense()` deliberately exempts `natively_api` from the HWID requirement — those licenses are server-validated per request, not device-bound. The three read paths did not agree with it: `readStoredLicense()`, `isPremium()` and `getLicenseDetails()` each bailed out on `!getHardwareId` before ever looking at the provider.

Consequence on any machine where `loadNativeModule()` returns null — an unbuilt dev checkout, an ASAR unpack failure, or AV quarantine of the `.node` on Windows — an API-plan key activated successfully and wrote `license.enc`, yet every read reported `isPremium: false`. The Modes Manager only *appeared* to unlock, because `App.tsx` trusts the `license-status-changed` event payload directly; Profile Intelligence queries `licenseGetDetails()` and so kept showing its Pro gate wall. Nothing in main ever agreed the user had Pro.

**Invariant the fix must not break:** gumroad and dodo licenses are HWID-bound, so with no native module their device binding is unverifiable and they must still resolve to `false`. Both branches are asserted in the new tests.

## Why this is not stacked

It touches only the gitlink and two brand-new files — nothing any of the seven stacked PRs goes near. This is a silent Pro revocation affecting paying API-plan users; making it wait on a 12-commit Windows stealth review would be the wrong trade. It applies cleanly straight onto `main`.

## Cross-platform analysis

- **Expected macOS behavior:** identical. The entitlement logic has no `process.platform` branch.
- **Expected Windows behavior:** identical. The native module is equally absent on both platforms in every failure mode this covers, and the provider branching is the same.
- **Impact radius:** entitlement reads — `isPremium()`, `getLicenseDetails()`, `readStoredLicense()`, and the `license-status-changed` payload consumers.

## Validation

- `Covered by automated macOS branch tests` — no platform branch, so one run covers both. 13 pass / 0 fail.
- `Covered by automated Windows branch tests` — same suite, executed on Windows 11.
- `Tested physically on Windows`
- `Requires physical macOS verification` — an end-to-end activation flow on macOS was not performed.

## Commands executed

```
npm run build:electron                                                    # ok
node --test electron/services/__tests__/LicenseNativeModuleAbsent.test.mjs   # 13 pass / 0 fail
```

Run at this exact branch tip, based directly on `origin/main` with no stack beneath it.

## Remaining risks

The real-world triggers — ASAR unpack failure, AV quarantine of the `.node` — are simulated in the test rather than reproduced. macOS activation was not exercised end to end.

https://claude.ai/code/session_01UWCt3EHyDEzRYhZXJwudSf

### DIFF
--- a/electron/ipcHandlers.ts
+++ b/electron/ipcHandlers.ts
@@ -6539,9 +6539,24 @@ export function initializeIpcHandlers(appState: AppState): void {
               if (!win.isDestroyed())
                 win.webContents.send('license-status-changed', { isPremium: true });
             });
-          } else if (result.skipped) {
+          } else if (result.skipped && !result.error) {
+            // Benign skip: the stored perpetual license is verified and already
+            // granting Pro, so there is nothing to tell the user.
             console.log(
               '[IPC] set-natively-api-key: existing Gumroad/Dodo license preserved — Pro not overwritten.',
+            );
+          } else if (result.skipped) {
+            // Skip WITH an error: the perpetual license could not be verified
+            // (native module absent), so it grants nothing AND the API key was
+            // refused — the user has Pro from neither credential. This is a
+            // support-visible fault, not the benign case above; do not bury it
+            // at log level. NOTE: still not surfaced in the UI — the
+            // license-status-changed payload is typed { isPremium, plan? } and
+            // has no channel for a reason string. Users hitting this via the
+            // license box DO see it (activateLicense returns result.error).
+            console.warn(
+              '[IPC] set-natively-api-key: Pro inactive —',
+              result.error,
             );
           } else {
             console.log('[IPC] set-natively-api-key: Pro not activated —', result.error);
--- a/electron/services/__tests__/LicenseNativeModuleAbsent.test.mjs
+++ b/electron/services/__tests__/LicenseNativeModuleAbsent.test.mjs
@@ -166,6 +166,22 @@ describe('native module absent: activateWithApiKey must not clobber a perpetual 
       assert.equal(result.success, false);
       assert.equal(result.skipped, true, `${provider} license must be preserved, not overwritten`);
       assert.deepEqual(fs.readFileSync(LICENSE_PATH), before, 'license.enc was modified');
+
+      // The skip MUST carry a reason here. Without a native module this stored
+      // license grants nothing (readStoredLicense rejects it), so refusing the
+      // API key leaves the user with Pro from neither credential. A bare
+      // `skipped` made activateLicense() render the empty error as "Failed to
+      // activate with API key" — blaming the one credential that is fine — and
+      // made the set-natively-api-key handler show nothing at all.
+      assert.ok(
+        result.error,
+        `${provider}: an unverifiable license must explain why Pro is inactive`,
+      );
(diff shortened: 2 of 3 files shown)


evinjohnn:
Pushed `bdd7c4b` to this branch — flagging since it changes the submodule pointer.

**Why:** this branch pinned `premium` at `69da127`, the tip of natively-premium#2 *before* review. That PR has now merged as `951e29f`, which includes a fix found during review that this branch's tests did not cover. Merging at `69da127` would have shipped the regression even though premium main was already fixed.

**The regression:** `activateWithApiKey` returned a bare `{ skipped: true }` for every stored Gumroad/Dodo license, conflating two opposite states. With the native module present the perpetual license *is* granting Pro, so skipping costs nothing. With it absent — the case this branch exists for — `readStoredLicense()` rejects that same license, so it grants nothing, and refusing the API key too leaves a user who paid for **both** credentials with Pro from neither. Since `skipped` carried no `error`, `activateLicense()` (LicenseManager.ts:137) rendered it as *"Failed to activate with API key"* — blaming the one credential that is fine. Via `set-natively-api-key` it was worse: logged, never shown.

**In this commit:**
- `premium`: `69da127` → `951e29f`
- `electron/ipcHandlers.ts`: branch on `error` so the real fault is warned rather than buried alongside the benign *"you already have a lifetime license"* skip
- `LicenseNativeModuleAbsent.test.mjs`: assert the skip carries a reason, and that the reason does not blame the API key

**Verification:** the two new assertions fail against `premium@69da127` (2 fail / 11 pass, both \`an unverifiable license must explain why Pro is inactive\`) and pass against `premium@951e29f` (13/13). They pin the fix, not the shape. `tsc -p electron/tsconfig.json --noEmit` clean.

**Still open, not addressed here:** no UI surface for the `set-natively-api-key` path — `license-status-changed` is typed `{ isPremium, plan? }` with no channel for a reason string, so widening it is a renderer change. Users hitting this through the license box do see the message.

https://claude.ai/code/session_018FEPZ3g9HWKcmL4fn4UZMn

PR #428, opened 2026-08-04
Fix codex cli and startup crash

cinnamon1952:
## Summary
This PR addresses two critical bugs:

1. **Startup / Shutdown `ENOENT` Crash Loop:** The `.env` template previously populated `GOOGLE_APPLICATION_CREDENTIALS` with a dummy path (`/path/to/your/service-account.json`). The `CredentialsManager` persisted this fake path to the local keychain/database, causing the app to repeatedly throw fatal `ENOENT` errors on `before-quit` or STT initialization even after removing it from the `.env` file. This PR adds a guard in `main.ts` to actively ignore and purge that specific dummy path from the credentials store on boot.
2. **Codex CLI Vision Support (Not an image model error):** The `CodexCliService` was incorrectly formatting image screenshot payloads using `{ type: 'input_image' }`. This caused the vision endpoint to silently drop the image, leaving models like `gpt-5.6-terra` to reply with "I am not an image model". This PR updates the payload to use the standard Chat Completions vision format (`{ type: 'image_url', image_url: { url: ... } }`).

Fixes # 

## Type of Change
- 🐛 Bug Fix
- [ ] ✨ New Feature
- [ ] 📝 Documentation

## Testing & Environment
- [x] Manual test performed on: **macOS (Apple Silicon)**
- Verified that the app no longer throws an `ENOENT` fatal crash when `GOOGLE_APPLICATION_CREDENTIALS` was previously set to the dummy path.
- Verified that taking a screenshot with the Codex CLI correctly sends the image payload and receives a vision-based response from the model instead of a fallback refusal.

### 1) A text only prompt
image

### 2) A prompt with screenshot attached
image


evinjohnn:
Closing — superseded by #435, which is now merged as `1b1453fd`.

Thanks @cinnamon1952 — both bugs you identified were real, and your diagnosis of each was correct:

**Codex CLI vision.** You were right that `buildRequestBody` was dropping `imagePaths` entirely and that this was why the model replied "I am not an image model". Your fix (`type: 'input_image'` with a data-URL string) is the correct Responses-API shape and shipped as-is. Worth recording for anyone who touches this later: the Chat Completions shape `{ type: 'image_url', image_url: { url } }` is **silently ignored** by `/v1/responses` — no error, the model just answers as though no image were attached. There's now a test pinning the string form specifically so it can't drift back.

**Startup ENOENT crash loop.** Also a real bug, and the `.env.example` placeholder was indeed the source. #435 generalised the fix: rather than matching the one dummy literal, it validates the key file and distinguishes a *definite* rejection ("this JSON is an OAuth client secret") from a transient one ("the volume isn't mounted"), because an unmounted `/Volumes` path reports ENOENT identically to a deleted file — evicting there would destroy a working credential. `.env.example` now ships the line commented out, so fresh clones no longer reproduce it.

What else went into #435 on top of your work:
- sharp downscale + size floor/cap on Codex images, matching the other cloud providers
- fail the turn when no image encodes, instead of silently answering screenshot-blind
- `CredentialsManager`: never overwrite a keyring file that failed to decrypt, and refuse degraded writes *before* mutating in-memory state (a Greptile P1 on #435)
- 41 tests, all mutation-probed

Two notes on this branch if you build on it again: it had drifted ~10 commits behind `main`, and the `AccordionSection.tsx` comment edit was unrelated to either fix — easier to review when a PR stays scoped to its own changes.

Appreciate the report and the fix — the vision payload one in particular was a genuinely non-obvious failure.

PR #436, opened 2026-08-05
ci: fetch the private premium submodule so Build Smoke can actually run

evinjohnn:
Every run of Build Smoke since it was added has died at its first real step:

```
electron/services/resolveCompanySearchProvider.ts(11,37): error TS2307:
Cannot find module '../../premium/electron/knowledge/CompanyResearchEngine'
```

`premium` is a **private submodule** both jobs genuinely need — `electron/tsconfig.json` includes `../premium/electron/**/*.ts`, and esbuild bundles `require('../premium/…')` calls — and `actions/checkout` was never asked to fetch it. So the **Windows leg added alongside it has never validated anything**, which is exactly the gap that let this stack's Windows defects through.

## The obvious fix does not work here

`submodules: true` runs `git submodule update --init --recursive`. This tree carries **two** gitlinks while `.gitmodules` describes only one:

```
$ git ls-tree origin/main | grep 160000
160000 commit 94edf288…  natively-api
160000 commit b591966f…  premium
$ cat .gitmodules        # premium only
```

Reproduced against a synthetic repo of that exact shape:

```
$ git submodule update --init --force --recursive
Submodule 'good' registered for path 'good'
fatal: No url found for submodule path 'orphan' in .gitmodules
   -> good populated?            # empty
```

It aborts **before** cloning the valid submodule. So `submodules: true` would fail the checkout outright even with a correct token. The workflow header recorded that plan; this PR records why it is wrong.

## What this does instead

A path-limited step — `git submodule update --init --force -- premium` — which the orphan cannot affect. Same repro:

```
$ git submodule update --init --force -- good
Submodule path 'good': checked out 'bb90c19…'
   -> good populated? f
   -> orphan left alone? 0 entries
```

**Auth:** `GITHUB_TOKEN` cannot read another private repo, so the step uses `SUBMODULE_TOKEN`, falling back to the existing `GH_APP_TOKEN`. The token is injected via a repo-**local** `url.insteadOf` rewrite and unset immediately after, so it never lands in global config. Availability is detected in a separate step because a step's own `env:` is not in scope for its own `if:` — the secret is never referenced in a condition.

**With no token the step is skipped, not failed.** The job then dies at the typecheck exactly as it does today — no worse than the status quo — and a `::warning::` names the reason. That matters because I cannot confirm from here whether `GH_APP_TOKEN` grants read access to `natively-premium`; if it does not, add a PAT as `SUBMODULE_TOKEN`.

## Not fixed here

The undescribed `natively-api` gitlink. Either it wants a `.gitmodules` entry or it should not be a gitlink — both are decisions about that repo's intent rather than about CI. Worth noting `git submodule status` currently fails on it for every developer too:

```
fatal: no submodule mapping found in .gitmodules for path 'natively-api'
```

## Validation

* `Build validated on macOS` / `Build validated on Windows` — **this PR's own run is the test.** The workflow triggers on `pull_request`, so the matrix will report on both legs. If the premium step succeeds and the typecheck passes, the fix works.
* YAML parses; both jobs get the steps in the intended order (verified with a parser, not by eye).
* The recursive abort and the path-limited workaround were each reproduced locally against a repo built to the same two-gitlink shape.

### Remaining risks

If `GH_APP_TOKEN` lacks access to `natively-premium`, the step skips and CI stays red exactly as now — visible in the step list rather than silent. The `natively-api` gitlink remains a latent trap for anything that later tries a recursive submodule operation.

### DIFF
--- a/.github/workflows/build-smoke.yml
+++ b/.github/workflows/build-smoke.yml
@@ -8,13 +8,14 @@ on:
   # since it was added completed with zero jobs, so no PR has actually been
   # smoke-tested. Verified: three most recent runs each report total_count 0.
   #
-  # Moving it makes the workflow PARSE; it does not yet make it PASS. Both jobs
-  # need the private `premium` submodule (electron/tsconfig.json includes
+  # Moving it made the workflow PARSE. Making it PASS additionally required the
+  # private `premium` submodule (electron/tsconfig.json includes
   # '../premium/electron/**/*.ts', and esbuild bundles require('../premium/…')
-  # calls), which actions/checkout cannot fetch with the default token. Until a
-  # PAT with read access to natively-premium is added as SUBMODULE_TOKEN and the
-  # checkout steps get `submodules: true` + `token:`, expect the jobs to fail at
-  # a named step — which is visible and actionable, unlike a run with zero jobs.
+  # calls) — see the checkout steps below, which fetch it explicitly.
+  #
+  # Note the earlier plan recorded here (`submodules: true` + `token:`) does NOT
+  # work in this repo: the tree has an undescribed `natively-api` gitlink that
+  # makes the recursive submodule init abort before `premium` is ever cloned.
   schedule:
     - cron: '0 9 * * 1' # Every Monday at 09:00 UTC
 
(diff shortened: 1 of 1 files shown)


evinjohnn:
## CI result: the fix works — Windows is green for the first time

```
Electron entry smoke (windows-latest): SUCCESS   ← all 17 steps
Electron entry smoke (macos-latest):   failure   ← at "Run Electron unit tests" only
```

Every build step passes on **both** legs now, including the one this PR exists for:

```
4. Checkout premium submodule: success
8. Typecheck Electron main process: success
```

That step had never once succeeded before.

### The remaining macOS failure is pre-existing, not from this PR

`npm test` is macOS-only in this workflow, which is why Windows is fully green and macOS is not. I ran the same command on plain `origin/main` with nothing from this branch applied:

```
$ npm test          # on origin/main @ 2866c856
574 distinct failing test names
```

and confirmed the CI-reported names are in that set:

```
PRE-EXISTING on main: curl provider save handlers validate renderer payloads before persistence
PRE-EXISTING on main: AppState sends model-changed only to current model listeners
PRE-EXISTING on main: the shadow block exists and is isolated correctly
PRE-EXISTING on main: main acquires the single-instance lock before native ABI recovery can rebuild or relaunch
```

This PR changes one file — `.github/workflows/build-smoke.yml` — and touches no application code, so it cannot have caused them. They were simply never visible before, because the workflow died at the typecheck long before reaching the test step. Making CI reach these suites is the point of this change; fixing them is separate work.

### A correction on the way here

The first push used `git config --local` for the token rewrite and failed:

```
fatal: could not read Username for 'https://github.com'
```

The submodule clone is a **subprocess** and does not inherit a `--local` rewrite. Reproduced both directions locally: with `--local` a deliberately bogus rewrite target was ignored entirely and the clone succeeded; with `git -c` the same bogus target failed, proving it reached the child via `GIT_CONFIG_PARAMETERS`. Fixed in `git -c`, which also keeps the token out of `.git/config` entirely.

`GH_APP_TOKEN` turned out to have the access needed, so no new secret is required.

PR #446, opened 2026-08-11
fix(context-os): a refusal pack only governs a bounded-universe mode

evinjohnn:
## The bug (user-reported, reproduced from their own trace)

What-To-Answer (Cmd+1 / Cmd+Enter) with a **screenshot** in a looking-for-work mode answered, three times:

> This is not directly mentioned in the uploaded material.

The user's `[CONTEXT-OS]` trace showed the whole chain: `sourceAuthority=profile_only`, `questionPreview=""` (screenshot only), `screen_context` in **forbiddenSources**, `selectedEvidenceCount=0` — and **`finalAction:"answer"`**. The kernel wanted to answer; the evidence layer overrode it. With zero candidates the coordinator pack comes back `refuse_insufficient_evidence`, both govern sites set `govern: true` **unconditionally**, and the surface yields the canned string without ever calling the model. The identical screenshot pasted into manual chat answered normally.

This is the same defect class fixed in `prompt-composer.ts` on 2026-08-07 — refusal keyed purely on "no evidence retrieved" — living a second time in Context OS, on the one surface that never goes through `buildV3Prompt`.

## The line

A refusal is truthful only when the mode's authority promises a **bounded universe** — exactly the four authorities `modeSourceContract` marks `evidenceRequired: true`. Everywhere else, an empty pack means the evidence system has nothing to add: fall back to the model.

`refusalPolicy.ts` is that line as a pure predicate, wired at both coordinator govern sites. `govern: false` reverts to the legacy path every consumer already guards on. **Doc-grounded refusals are untouched** — a `funding_source` question against a real uploaded paper still refuses. Unknown authorities fail toward answering. A drift-guard test derives the real `evidenceRequired` mapping from `defaultSourceContractForNewMode` and asserts the mirror cannot drift.

Also: the refusal string now takes the pack's `sourceOwner`, so a profile-owned refusal — if ever yielded — says "your profile material" instead of falsely claiming uploads (stem preserved for `REFUSAL_SNIFF_RE`). The `ContextOsStaticImport` tripwire is **sharpened**: it now asserts the contract-denial clear (the leak defense) stays unconditional and the coordinator de-dup clear stays govern-gated, instead of a bare line count that conflated the two.

## Verification

- new suite **16/16** (predicate matrix, drift guard, string honesty)
- WTA/Context-OS/evidence suites **53/53** on this branch
- e2e sim of the exact reported turn against compiled code: `profile_only` refusal → `govern:false` → model answers; `reference_files_only` → still refuses
- attribution runs (clean HEAD vs HEAD + only these files, Electron runner): llm glob **18 = 18** failing names, services glob **286 = 286** — zero new failures at name level
- `tsc` clean

## Assembly note

Built from `origin/main` directly. The shared working tree carried another agent's unmerged commit in `IntelligenceEngine.ts` and **uncommitted LiteLLM reversions in `ipcHandlers.ts`** — both files were rebuilt from main's versions with only this fix's hunks applied, verified hunk-by-hunk (`optInFamily`/`litellmPreferredModel` intact).

## Known, deliberately out of scope

- The §6 Answer-policy setting is consulted **nowhere** in `context-os/**` — it silently does nothing on the WTA surface. Wiring it is a subsystem change; this restores the correct default first.
- `required: profileKinds()` still demands both `profile_resume` **and** `projects` on every `profile_only` turn (the starved-family refusal the review found). Moot for user-visible refusals after this fix, but still shapes evidence selection.
- The `ask_clarification` path still yields a developer-diagnostic string (`contract.reason`).

### DIFF
--- a/electron/IntelligenceEngine.ts
+++ b/electron/IntelligenceEngine.ts
@@ -2452,6 +2452,14 @@ export class IntelligenceEngine extends EventEmitter {
             if (!wtaContextOsGeneration
                 && wtaTurnContract
                 && documentGroundedCustomModeActive
+                // Live proof 2026-08-11: documentGroundedCustomModeActive can be
+                // TRUE with hasReferenceFiles FALSE (custom mode, contract seeded
+                // reference_files_primary, zero files). Governing that turn
+                // builds an empty pack whose answerPolicy is ask_clarification,
+                // and the surface then yielded contract.reason — the raw
+                // developer diagnostic — to the user. A doc-grounded govern
+                // with no documents governs nothing: require the files.
+                && Boolean((snapshotModeInfo as any)?.hasReferenceFiles)
                 && isIntelligenceFlagEnabled('contextOsEvidencePackEnabled')) {
                 wtaContextOsGeneration = {
                     contract: wtaTurnContract,
--- a/electron/IntelligenceEngine.ts
+++ b/electron/IntelligenceEngine.ts
@@ -905,6 +905,13 @@ export class IntelligenceEngine extends EventEmitter {
         // every live token (#3) so the renderer can drop stale-generation batches.
         const snapshotModeInfo = this.getActiveModeInfo();
         const documentGroundedCustomModeActive = snapshotModeInfo?.documentGroundedCustomModeActive === true;
+        // Defect C split (2026-08-01): STRICT knowledge suppression vs broad
+        // source isolation. Strictness consumers below (skip-legacy-retrieval,
+        // forceDocumentGrounding, the generic-knowledge bypass gate, and the
+        // doc-grounded govern site) read THIS; isolation consumers keep the
+        // broad flag. WTA was never migrated when manual chat was
+        // (LLMHelper:5448) — the root asymmetry behind the 2026-08-11 reports.
+        const strictDocumentGroundedActive = (snapshotModeInfo as any)?.strictDocumentGroundedActive === true;
         const snapshotModeId = this.getActiveModeId();
         // The narrow ActiveModeInfo snapshot is enough for planning, but a
         // multi-family typed reference pack also needs the full mode row and its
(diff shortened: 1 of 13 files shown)


evinjohnn:
Pushed a second commit: the same live test surfaced a **third** denial path — the legacy `sourceOwnership` clarification reached via the Context OS clarify short-circuit, firing in a blank general-template mode (seeded `reference_files_primary`, zero files, empty question, screenshot). The first turn in the same log confirms the original refuse-path fix works (320-char vision answer on the exact original repro). New predicate `clarificationIsActionable` gates both short-circuits; WTA also gains the `!imagePaths` bypass the manual twin always had. 24/24 + 67/67, tsc clean.

evinjohnn:
## Review follow-up (2026-08-12): findings F1–F3 + F8 fixed in 14b937d6

The independent review verdict was **CONDITIONAL** — the reported keystroke was closed, but not the symptom class. Now addressed:

- **F1 (HIGH, fixed)** — `IntelligenceEngine` post-answer doc validator was the *seventh* strictness consumer still reading the broad `documentGroundedCustomModeActive` flag. One typed question in the reported fileless mode would have overwritten a correct streamed answer with *"I could not find that in the retrieved sections of the document."* Gate now requires strict flag + `hasReferenceFiles` + `isDocGroundedAnswerType` (the manual twin's existing parity terms).
- **F2 (HIGH, fixed)** — `packGovernsGeneration` now takes `hasReferenceFiles`: the reference-bound trio only forms a bounded universe when files exist; omitted fails toward answering; `transcript_only` stays authority-only. Both govern call sites pass it.
- **F3 (MED, fixed)** — WTA `governedEvidencePack` now keys on `.govern`, not pack presence, so an ungoverned refuse pack no longer suppresses legacy mode-context retrieval.
- **F8 (fixed)** — comment indentation.

**Deferred, deliberately:**
- **F4** — `strictDocumentGroundedActive` narrows isolation for template-seeded modes *with* files via `plan.documentGroundedCustomModeActive` consumers (contextRoute / conversationHistoryPolicy). Review could not construct an actual leak; second-layer defenses hold. Tracked, not blocking.
- **F5** — `AnswerPlanner`'s `strict ?? broad` fallback is a footgun for callers that omit the strict flag (flips `unknown_answer`→`lecture_answer`); no live caller omits it today.
- **F6** — the §6 Answer-policy setting is still consulted nowhere in strictness derivation — **pre-existing**, and the remaining open wiring gap of the original request.
- **F7** — the authority drift guard only exercises 2 template-derived authorities; near-vacuous, needs a widened fixture.

Tests written red-first (27/31 → 31/31 on the F2 matrix); suites green: 101/101 on this branch, 110/110 in the integration tree.

PR #454, opened 2026-08-12
fix(answers): [[GIST]] split, doc-grounded flag split, coding-repair guards

evinjohnn:
Opening this for the branch as it stands. **It carries 25 commits from several agents**, not just the one described below — the branch is also 31 commits behind `main`, so it needs a merge or rebase before it can land.

### `[[GIST]]` display markup (renderer + main-process twin, kept in lockstep)
The reported symptom: the marker rendering as literal text above its essence.

- **Recover a marker whose essence the model put on the next line.** The old split rejected the whole shape, so a literal `[[GIST]]` painted on screen. Recovery is bounded — exactly one following line, at most ten words (the prompt contracts for five to eight). A longer line is likelier a real closing sentence under a misplaced marker, and demoting it to a chip would silently drop it from the *spoken* answer.
- **A line-leading marker never survives into the body**, even with no essence after it. `stripDisplayMarkup` feeds TTS, which would otherwise say it aloud. (This one was a regression in my own first fix, caught by review.)
- **The streaming split hides a completed marker awaiting its essence**, so it no longer flashes between `[[GIST]]\n` and the first essence token.
- `spokenFormatViolations` still reports a recovered marker as `gist_misplaced` — the recovery is containment, and the lint is how the drift stays measurable.

### Document-grounded mode — one flag was answering two incompatible questions
Each past narrowing of it broke the other role.

- **Isolation** (résumé/JD/Hindsight/OKF suppression) is authority-only *by design* since 2026-07-15 and must engage before the first upload. Keying it on the strict flag — false for every template-seeded mode — silently re-opened résumé injection on document-grounded turns.
- **Routing and the prompt policy line** now require `documentGroundedFilesPresent`: with zero files they describe a universe that does not exist, which is what routed general questions into doc shapes that then refused against an empty evidence block.

Turns the branch's red `DocGroundedRoutingRequiresFiles2026_08_05` spec green (7/7).

### Coding validator/repairer
- `usesCanonicalHeadings` now needs a real `##` heading *outside fences* **and** a recognized section. `sectionHeader` deliberately matches a bold `**Approach**` for completeness checks, so the widened test caught the model's own format and fed complete bold-heading answers back into the lossy repair — re-opening the 2026-08-10 regression the exemption exists to prevent. Answers that started the canonical scaffold and dropped out are still repaired.
- The repairer strips fences before its last-resort complexity scan. A Big-O in a *discarded-approach* comment was being published as the shipped code's official Complexity section.

### Refusal wording
Only `'profile'` was special-cased, so transcript- and mixed-owned refusals still claimed to have searched "the uploaded material" in modes with no uploaded files. Per-owner stems now; every stem keeps the "not directly mentioned" opening `REFUSAL_SNIFF_RE` needs.

### Governed-pack gate
The post-stream doc-grounded validator keyed on pack *presence*, but a pack rides along even when it does not govern. Such a pack has no items, so the validator checked the answer against an empty evidence block and refused a turn the legacy path had answered.

## Verification of `77bb5b14`

From a **clean detached worktree**, not the shared working tree:

| | |
|---|---|
| `build:electron` | clean |
| `tsc --noEmit` | exit 0 |
| 4 new test files (26 tests) | all pass |
| `test:lib` | 317/317 |
| `electron/llm` suite vs **its own parent commit** | identical failing-name diff — **zero regressions** |

The coding tests were confirmed to fail against the pre-fix validator (4 of 7) before being kept, so they aren't vacuous.

## Caveats

- The other 24 commits on this branch are other agents' work and were not reviewed or verified by me.
- 14 pre-existing failures in `electron/llm` and 3 in intelligence are present on the branch, unrelated to this commit and untouched.
- The prompt-wording half of the `[[GIST]]` fix is unverified against a live model; the parser change is what's load-bearing.
- Nothing was run in the packaged app or on Windows. All changes are platform-independent string/flag logic.

https://claude.ai/code/session_01PSXKVJvchN6bPJK2fKT4Rw


evinjohnn:
## Answer-pipeline workstream — summary of 22 commits on this branch

This branch carries work from several agents. This comment covers only the answer-pipeline commits (`e62e8524` … `c2ad3133`); the `[[GIST]]` work in the PR description is separate.

### The headline fix: a second answer was being appended to the first

The long-standing report — *"it shows a new answer that's wrong after initially showing the right answer"* — was in the **client**, not the `natively-api` cascade. Proved from log arithmetic before touching code:

```
stream 1  tokens 8047  chars 22871  -> ai_unavailable at 61s
stream 2                chars  2342  <- a whole new answer
stored assistant message      25210
22871 + 2342 = 25213 vs 25210 stored -> a 3-char trim delta = CONCATENATION
```

`yield*` cannot be undone. Every catch-and-continue site fell through on **any** error, including errors raised long after tokens had reached the renderer. The rule was already written down for the vision path (*"A failure AFTER commit cannot switch providers (that would duplicate output)"*) and implemented there — the text path documented it and never applied it at its own sites. Now guarded at **8 sites** via `trackCommit`.

### Other user-visible fixes

- **Runaway generation** — both existing guards are *time*-based (13s to first token, 8s inter-token stall) and a runaway is fast, so neither could ever fire. Added `MAX_STREAM_OUTPUT_CHARS`, sized off measured data (largest real answer observed: 2530 chars). Server-side counterpart: `evinjohnn/natively-api#4`.
- **Silent truncation** — a stream ending early was indistinguishable from a complete one, so truncated answers were stored as conversation history and became the antecedent for the *next* turn's referent resolution. Fixed on **both** manual chat and WTA.
- **`$@` / `$<` rendered as KaTeX** — shell and Make variables parsed as inline math, mangling the question echo. Fixed via the pandoc adjacency rule, a shell-sigil rule, and escaping non-math `$` so `remark-math` cannot reclaim them.
- **`(referring to: Makefile)`** — `its` in "on its own line" triggered referent inheritance, gluing an unrelated topic onto a self-contained question and driving retrieval with it.
- **Coding contract lost on V3 turns** — the six-section contract never reached the model, on live *and* manual-chat coding questions.
- **Test suite 426 → ~123 failures** via two systemic fixes: restored a fixture corpus deleted by a gitignore sweep, and `npm test` now uses the Electron runner (native modules are Electron-ABI on purpose).

### Verification

7,343 tests. The failing set matches the pre-existing baseline plus one known order-dependent test (`IntelligenceEngineScaffoldContaminationFallback`), bisected and confirmed to fail standalone **without** any of these changes. Live-verified against `api.natively.software` throughout; several defects were found only by live runs after passing offline tests. Every fix is red-green verified.

### Caveats

- **macOS only.** All changed code is platform-independent TypeScript with no `process.platform` branches — Windows is *reviewed, not executed*.
- **~123 pre-existing failures remain** (Codex sandbox, Keychain entitlements, unset-URL suites, ONNX flakes). Unrelated and untouched.
- **The duplication and runaway guards have no live evidence** — they only fire when a provider misbehaves. Confirmed not to break anything; not confirmed to fire in production.
- Three live checks remain unrun, most importantly a coding question through **manual chat**.
- `9b78621b` inadvertently includes two hunks from another session's uncommitted work in `ipcHandlers.ts` (a `hasReferenceFiles` parity fix and a `_governedPack` re-key). Two reviewers independently confirmed both are correct; they rode in unreviewed by me. Flagged rather than rewritten, since rewriting shared history risks destroying in-progress work.

### Blocker unchanged

The branch is **33 behind / 26 ahead** of `main` and still needs a merge or rebase before it can land.

Three rounds of multi-agent review ran against these commits; each round found something the previous missed, and all findings are closed.

evinjohnn:
@greptile-apps review

This merged with no automated review — the bot skipped it at 119 files (100-file limit) and nobody tagged it. It carried a full answer-pipeline workstream (streaming duplication guard, runaway output cap, truncated-answer storage, markdown `$` escaping, referent resolution) plus commits from three other agents in a shared working tree.

Worth a retroactive pass on `electron/LLMHelper.ts`, `electron/ipcHandlers.ts`, `electron/llm/WhatToAnswerLLM.ts`, `electron/IntelligenceEngine.ts`, `electron/context-intelligence/question/conversation-state.ts` and `src/lib/streamingMarkdown.ts` in particular — those hold the behavioural changes.

PR #464, opened 2026-08-14
feat(llm): test-only switches to provoke the stream guards

evinjohnn:
## Why

Two guards added 2026-08-12 only fire when a provider misbehaves:

- the **post-commit guard** (`trackCommit`) — a provider failing *after* its first token must end the turn rather than fall through, which used to append a second, different, complete answer to a partial one
- the **runaway output cap** (`capOutput`)

Neither could be summoned on demand, so their honest status was *"confirmed not to break anything; not confirmed to work"* — a bad place for guards whose entire job is a rare failure.

## What

Two dev-only switches:

```bash
NATIVELY_TEST_FAIL_STREAM_AFTER_CHARS=200   # provider dies mid-answer
NATIVELY_TEST_STREAM_OUTPUT_CHARS=400       # lower the output ceiling
```

The fault throws from inside `trackCommit` **after** the yield, deliberately: the case worth demonstrating is a provider dying once output is already on screen.

Verified locally: with the switch at 40, the stream throws `InjectedStreamFault` at 42 chars with `committed = true` — exactly the state the guard keys off. The ceiling switch fires at 64 > 60 and reaches **both** cap sites; reaching only one would make the other look bounded when it wasn't.

## Safety

These are product code now, so a fault injector that could activate in a shipped build would be worse than the bugs it demonstrates. **Three independent conditions, all required:**

1. an explicit env var (never on by default)
2. a **non-packaged** build — `app.isPackaged === false`
3. a value that parses sane

`0` and negatives are rejected: a zero-char fault fires *pre*-commit, which already fails over correctly and would demonstrate the wrong thing; a zero ceiling would truncate every answer to nothing. Electron is resolved lazily so pure-logic tests can still import the module.

8 tests, including one that walks **every** exported switch and fails if a new one is added without the packaging gate.

## Notes for review

- This was authored on a branch that also carried an unmerged **per-surface ceiling** (`MAX_SUMMARY_OUTPUT_CHARS` / `long_form`). That does not exist on `main` — neither the constant nor the `profile` parameter — so the conflict was resolved against `main` and that work was deliberately **not** carried across. A companion commit that only made sense on that branch was dropped.
- Cherry-picked onto a fresh branch off `main` via a worktree, so another session's in-flight changes in the shared checkout were never disturbed.
- Full-suite execution in the worktree is blocked by `premium` being checked out on a different branch and referencing modules absent from `main` — environment skew, unrelated to this change. The fault-injection suite passes 8/8; please let CI run the rest.

### DIFF
--- a/electron/LLMHelper.ts
+++ b/electron/LLMHelper.ts
@@ -5161,8 +5161,13 @@ let isMultimodal = !!(imagePaths?.length);
     // return unconditionally (Ollama, OpenAI, Claude, DeepSeek, LiteLLM) alike.
     // See MAX_STREAM_OUTPUT_CHARS for why this is a character cap and not a
     // wall-clock one, and why it is defence in depth rather than the real fix.
-    const { MAX_STREAM_OUTPUT_CHARS } = await import('./llm/liveDeadlines');
-    const { testOutputCharCeiling } = await import('./llm/streamFaultInjection');
+    // Loaded together: the two imports are independent, and awaiting them in
+    // sequence costs an extra microtask hop on every streamed turn for nothing
+    // (react-doctor server-sequential-independent-await).
+    const [{ MAX_STREAM_OUTPUT_CHARS }, { testOutputCharCeiling }] = await Promise.all([
+      import('./llm/liveDeadlines'),
+      import('./llm/streamFaultInjection'),
+    ]);
     // Test switch (dev-only, opt-in) lets the cap be provoked without waiting
     // for a model to actually loop. Null in any packaged build.
     //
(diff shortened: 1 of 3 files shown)


evinjohnn:
Closing as **already upstream**, not as rejected.

PR #466 swept this work into `main`. Verified:

- `electron/llm/streamFaultInjection.ts` — **zero diff** against `origin/main`; both switches (`failStreamAfterChars`, `testOutputCharCeiling`) are present.
- `electron/llm/__tests__/StreamFaultInjection2026_08_14.test.mjs` — **zero diff** against `origin/main`.

Merging it now would be actively harmful. The branch predates #466, so `git diff origin/main origin/fix/stream-fault-injection` reports **132 files changed, 9,817 deletions** — those "deletions" are main's newer work that this branch simply doesn't have yet.

The rebase made the regression concrete. In `LLMHelper.ts`, main now computes a **per-surface** ceiling:

```ts
const outputCeiling = testOutputCharCeiling()
  ?? (profile === 'long_form' ? MAX_SUMMARY_OUTPUT_CHARS : MAX_STREAM_OUTPUT_CHARS);
```

This branch's side reverts that to the single `MAX_STREAM_OUTPUT_CHARS` cap — its own in-code comment explains it was deliberately resolved that way because the per-surface ceiling did not exist at the time. It does now, so that resolution is stale and would cap whole-meeting summaries at the live-answer ceiling.

Nothing is lost by closing: the feature and its tests are on `main` today.

PR #466, opened 2026-08-14
Semantic-retrieval repair, TS7 migration, audit fixes, and review-findings closure

evinjohnn:
## What

57 commits across four workstreams that were built and validated together on this branch:

### 1. Semantic-retrieval repair (audit → phased fix → calibrated default-ON)
- Full cosine/retrieval audit (`reports/cosine-similarity-relevance-audit-2026-08-13.md`), then Phases 0.5–5: `semanticAdmissionGate` (admission ≠ ranking, calibrated gemini-768 floor 0.69 — live-verified that the legacy scorer retrieved ONLY a boost-carrying irrelevant node while rejecting all relevant ones), jd_fit V2 coverage, unified hybrid eligibility across both LLMHelper entry points (`modeHybridEligibility.ts`), space-keyed thresholds + `[SemanticAdmission]` telemetry, Windows `sqlite-vec-windows-x64` parity.
- Both flags are kill-switch DEFAULT ON, disableable at runtime without redeploy. Pairs with natively-premium PR #5 (**merge that first** — the premium gitlink here references its head).

### 2. TS7 migration
TypeScript 7 is the primary checker on CLI/CI (`typescript7` alias; TS5 retained for the pre-commit hook). Typecheck is 0 errors on both electron and renderer configs.

### 3. Audit fixes (F-102…F-201) + usage ledger
Route-change/recovery races, stream guards, outbox durability, gemini-3.1-flash-lite + 3.7-flash defaults, usage instrumentation.

### 4. High-effort code-review closure (`1ef70cbb`)
All 10 confirmed correctness findings fixed — ReviewModal consent (explicit decline path; 'Send anonymously' labeling), route-change error-path rollback, truncation surfaced to buffered callers and RAG consumers, outbox MAX_ATTEMPTS honoring its durability contract, `isBuiltin` case fix, ledger honesty on null answers, AnswerPlanner honesty for document-content questions, and the revived `mac-screen-recording-revoked-rebuild` diagnostic.

## Validation
- TS7 typecheck clean (electron + renderer); build clean.
- Deterministic: retrieval suites 26/26, premium tree 207/207, context-intelligence 905/905, Zerofill 21/21, UsageOutbox 18/18, ReviewModal contracts 19/19, TccRepair 11/11, stageCatalog 35/35. rag tree 256/260 (4 pre-existing fails, untouched baseline).
- Live: 22/22 E2E vs real DeepSeek + gemini-embedding-2; 13/13 flag matrix vs the real Natively backend (both flag states).
[...]

PR #468, opened 2026-08-15
fix(tests): make the remaining tsc-invocation helpers Windows-correct

evinjohnn:
Finishes the bug class fixed for the Context OS suites in the previous CI change, and closes the workflow's own TODO:

> `TODO: fix the Windows failures in these suites and drop both if: gates.`

## The defects

Ten test helpers build an isolated tsc tree so they can `cjsRequire` compiled output. Each had the same two Windows-only breakages:

1. **A mkdtemp path interpolated unquoted into a shell string** — `execSync(\`node ... --outDir ${target}\`)`. On a Windows runner `target` is `C:\Users\RUNNER~1\AppData\Local\Temp\...`.
2. **`node_modules/.bin/tsc` cannot be launched by a no-shell exec on Windows** — extensionless on POSIX, `tsc.cmd` on Windows.

Seven also passed `'dir'` as the symlink type; on Windows a directory symlink needs privileges a junction does not. Those now use `process.platform === 'win32' ? 'junction' : 'dir'` — the exact form the three already-correct helpers in this same set use.

All replaced with the form CLAUDE.md prescribes, already proven on a Windows runner: `execFileSync` with an args array, tsc as `node <entry>`, `path.join` throughout.

## Not a compiler migration

Each file keeps its own compiler and tsconfig. `.bin/tsc` resolves to typescript 5.x so it becomes an explicit `typescript/bin/tsc`; the three sites on `typescript7` + `tsconfig.emit.json` stay there. The emitted tree these tests require is unchanged.

## Validation

`Covered by automated macOS branch tests` — `npm test`, same worktree before and after:

| | pass | fail |
|---|---|---|
| main (`86612cd5`) | 7296 | 153 |
| this branch | 7312 | 137 |

**No new failures.** Two names appear only in the "after" set because the run now gets *further* — 16 more tests execute, surfacing `ModeSourceContractCreateThenUpdateBug2026_07_11`, a file this PR does not touch. Run in isolation it is **3 fail on clean main** and **1 fail here**, so this change improves it.

`Requires Windows CI verification` — Windows cannot be proven before merge: that leg is `continue-on-error`, and these suites' Windows failures are exactly what this fixes. Confirm after landing by diffing the Windows failing-name set, as the previous CI fix was confirmed.

## Coordination

`electron/services/__tests__/LLMHelperNegotiationCoachingGate.test.mjs` is also touched by #461, but in a different hunk — #461 does not modify the tsc call. #461 is already conflicting with main and needs a rebase regardless.

### DIFF
--- a/electron/llm/__tests__/AssistantRefusalDetection2026_07_23.test.mjs
+++ b/electron/llm/__tests__/AssistantRefusalDetection2026_07_23.test.mjs
@@ -17,7 +17,7 @@ import assert from 'node:assert/strict';
 import path from 'node:path';
 import os from 'node:os';
 import fs from 'node:fs';
-import { execSync } from 'node:child_process';
+import { execFileSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 import { createRequire } from 'node:module';
 
@@ -28,8 +28,12 @@ const distDir = (() => {
   const bundled = path.resolve(repoRoot, 'dist-electron/electron/llm/documentGroundedPrompt.js');
   if (fs.existsSync(bundled)) return path.resolve(repoRoot, 'dist-electron');
   const target = fs.mkdtempSync(path.join(os.tmpdir(), 'refusal-dist-'));
-  fs.symlinkSync(path.join(repoRoot, 'node_modules'), path.join(target, 'node_modules'), 'dir');
-  try { execSync(`node node_modules/.bin/tsc -p electron/tsconfig.json --outDir ${target}`, { cwd: repoRoot, stdio: 'pipe' }); } catch { /* expected partial */ }
+  fs.symlinkSync(path.join(repoRoot, 'node_modules'), path.join(target, 'node_modules'), process.platform === 'win32' ? 'junction' : 'dir');
+  try { execFileSync(process.execPath, [
+    path.join('node_modules', 'typescript', 'bin', 'tsc'),
+    '-p', path.join('electron', 'tsconfig.json'),
+    '--outDir', target,
+  ], { cwd: repoRoot, stdio: 'pipe' }); } catch { /* expected partial */ }
   return target;
 })();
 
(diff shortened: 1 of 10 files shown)


evinjohnn:
## CI result — and a correction to this PR's stated benefit

**Zero regressions, but also zero currently-failing tests fixed.** Both platforms are byte-identical to `main`:

| | main `86612cd5` | this PR |
|---|---|---|
| macOS | 138 fail / 165 names | 138 fail / 165 names |
| Windows | 177 fail / 212 names | 177 fail / 212 names |

`comm` reports an empty diff in both directions on both platforms.

**Why the predicted Windows win did not materialize:** I scoped this work from a `main` Build Smoke run that predated #466/#467. On today's `main`, `tsc emission failed — LLMHelper.js missing from isolated tree` occurs **0 times** on Windows — that failure was already resolved upstream. So the tests this PR touches were not failing on Windows any more, and making their invocations correct changes no outcome today.

**What this PR is, accurately:** preventative hardening, not a failure fix. It removes three real Windows portability defects from ten helpers — a shell-interpolated mkdtemp path, a `.bin/tsc` shim that a no-shell exec cannot launch on Windows, and seven `'dir'` symlinks that need privileges a junction does not. Those defects are latent right now; they break the moment these paths are exercised on a Windows host again, which is exactly how the Context OS suites broke.

One of the ten, `LLMHelperNegotiationCoachingGate.test.mjs`, still fails on Windows — **identically on `main` and on this branch**. It is an assertion failure, not an emission failure, and it sits among a broader `services/` Windows failure cluster (`LexicalFloorScaleF23`, `LexicalTokensNumerals`, and others). Out of scope here; not caused or fixed by this change.

Local macOS evidence is unchanged and still favourable: `npm test` goes 7296 pass / 153 fail → 7312 pass / 137 fail, because the run gets further.

PR #476, opened 2026-08-16
feat(stt): Nemotron 3.5 ASR Streaming as a local STT model

evinjohnn:
Adds **Nemotron 3.5 ASR Streaming** (NVIDIA cache-aware FastConformer-RNNT, `onnx-community/nemotron-3.5-asr-streaming-0.6b-onnx-int4`, 793 MB int4) as a local STT model.

It is the only model in the catalog with **real streaming** — chunked ONNX inference carrying encoder cache state across calls — rather than the simulated polling every existing entry (Whisper / Distil-Whisper / Moonshine / Parakeet) uses. It is also multilingual across 40 tiered locales.

34 commits, 58 files, +5,591 / −82.

## What's here

**New engine** — `electron/audio/whisper/nemotron/`
- `nemotronEngine.ts` — session management + chunk loop (encoder / decoder / joiner)
- `melFrontend.ts` — mel spectrogram frontend with real cross-chunk PCM lookback
- `rnntDecoder.ts` — greedy RNNT decode, `lastTokenId` carried across frame boundaries
- `tokenizer.ts` — SentencePiece detokenization with a `vocab.txt` fallback
- `languageTable.ts` — 40 locales with verified `lang_id` conditioning values
- `downloadFiles.ts` / `cacheState.ts` — flat-repo layout (no `onnx/` subdir, no dtype suffix), so it needs its own cache check rather than `expectedOnnxFiles()`
- `sharedWorkerRegistry.ts` — both audio channels (mic + system) served from **one** shared worker, with serialized message dispatch

**Wiring** — new `sessionLayout: 'nemotron-rnnt'` in `types.ts`, routed through `whisperWorker.ts`; a dedicated download provider in `LocalModelDownloadService.ts`; weighted ONNX slot acquisition in `onnxThreadConfig.ts` (three sessions per instance, two channels).

**Build** — `onnxruntime-node` externalized in `scripts/build-electron.js`.

## Why the model was hidden, and why it isn't now

An earlier commit set `hidden: true` on the catalog entry: real-model testing showed it transcribing speech as an **empty string** — 0% word overlap against whisper-tiny.en's 78–100% on identical audio. A complete functional failure, not a quality gap.

Two root causes, both fixed here:
1. The driving code baked synthetic zero-padding into the encoder's leading 9 `pre_encode_cache` frames instead of real cross-chunk mel history. `NemotronEngine` now carries a raw-PCM lookback buffer across chunk boundaries.
2. `lang_id` was 2947 — a vocab-token id, not a conditioning index. `DEFAULT_LANG_ID` is now 0, empirically verified.

The go/no-go integration test now passes at **77.8% word overlap** against a required 50%, so the entry is unhidden.

## Tests

- `integration.test.mjs` — real-model go/no-go against ground-truth audio
- `multilang.test.mjs` — transcription accuracy across 18 real languages (WAV fixtures included)
- `melFrontend.test.mjs` — golden values pinned for the 6 ground-truth mel constants
- `rnntDecoder.test.mjs`, `tokenizer.test.mjs`, `languageTable.test.mjs`
- `OnnxSlotWeightedAcquisition.test.mjs` — proves dual-channel slot weighting and that worker message serialization actually prevents overlap

## Cross-platform status

`onnxruntime-node` is a native module, and this PR changes how it is externalized at build time. Per the repo's cross-platform contract that is a flagged area.

- **Reviewed but not executed on Windows** — the engine is platform-neutral TypeScript over onnxruntime-node, but neither the packaged Windows build nor Windows EP selection has been exercised.
- **Requires physical Windows verification** — native-module loading from the packaged app, and the shared-worker path under two concurrent audio channels.
- EP-support findings are written up in `docs/superpowers/plans/nemotron-ep-support.md`.

## Notes for review

The branch was ~2 days stale against a main that had moved 140 commits; it test-merges cleanly (`git merge-tree` reports no conflicts). Highest-value review surface is the native-module / build-config / ONNX-slot-accounting side rather than the ASR math, which is covered by the golden-value and real-model tests.

https://claude.ai/code/session_011VSHtwCxXjMVdyheoF73Bd


evinjohnn:
@greptile-apps review

PR #481, opened 2026-08-17
fix(stt): make Nemotron usable — CPU+threaded inference (3x), preload (20x load), and no more mid-inference SIGABRT

evinjohnn:
Follow-up to #476. Nemotron produced no transcript and crashed the app; Parakeet was fine on the same machine, same pipeline. That contrast is what cracked it — **three defects, all integration, none in the model.**

---

# 0. Performance: 3× faster inference, 20× faster load

Nemotron felt far slower than Parakeet despite both being ~0.6B. Parameter count is the wrong axis: **Parakeet is CTC** (one session, one encoder pass per chunk); **Nemotron is RNN-T** (three sessions; per chunk an encoder pass *plus* a per-frame greedy loop where every emitted symbol costs a decoder run **and** a joint run). Published profiling of Parakeet RNN-T puts **~67% of runtime in greedy decoding vs ~33% in the encoder** — so the decode loop, not the encoder, is what to tune.

Two settings inherited from the Whisper-shaped default were hurting it badly. M-series (10 logical / 4 performance cores), int4 export, 2.46s fixture, median of 3 after warmup. **Transcript byte-identical in every configuration** — none of this trades accuracy:

| config | load | inference | RTF | ×realtime |
|---|---|---|---|---|
| coreml+cpu, 1 thread ← **was shipped** | 6436ms | 1158ms | 0.470 | 2.13× |
| cpu, 1 thread | 363ms | 936ms | 0.380 | 2.63× |
| cpu, 2 threads | 325ms | 547ms | 0.222 | 4.50× |
| **cpu, 4 threads ← now** | **326ms** | **385ms** | **0.156** | **6.39×** |
| cpu, 8 threads | 324ms | 528ms | 0.214 | 4.66× |

**CoreML is a straight loss on this model, in both directions** — ~6.1s of graph compilation at load *and* ~24% slower inference. The app's own startup log shows why: CoreML partitions the encoder into **369 fragments** (971/1891 nodes supported) and each 13-node decoder/joint graph into 3. Every boundary is a CPU↔CoreML copy, and the RNN-T loop crosses them *per symbol* rather than once per chunk. This is the [documented ORT failure mode for heavily-partitioned graphs](https://github.com/microsoft/onnxruntime/issues/17654), and it's why the [upstream reference implementation of this exact export](https://github.com/codavidgarcia/nemotron-3.5-asr-streaming-onnx) ships CPU-only. Nemotron now opts out of the platform accelerator everywhere (CoreML *and* DirectML); every other model keeps it.

**`intraOpNumThreads` was pinned to 1 for all consumers.** A CTC pass barely notices; the RNN-T loop pays it on ~two orders of magnitude more ORT invocations. Added an `OnnxWorkload` discriminator so **only** `'rnnt-decode'` gets a budget — every existing caller keeps `intra=1` byte-for-byte. The budget tracks cores but **caps at 4**, because 8 measured *worse* than 4: past the performance-core count the pool spills onto efficiency cores and contends. The cap is the measurement, not a guess.

`enableCpuMemArena` / `enableMemPattern` are deliberately left **off**. Enabling both measured 385ms → 374ms (~3%) — not worth re-introducing BFCArena growth, the documented mechanism behind the crash reports that gate exists to prevent.

---

The engine itself is healthy. Real-model go/no-go against the downloaded 754MB checkpoint:

```
Nemotron transcribed: "quick brown fox jump s over the la zy dog"
Word overlap: 0.778  ✔
```

---

## 1. Why it crashed — a teardown asymmetry

```
libc++abi: terminating due to uncaught exception of type Napi::Error
Electron exited with signal SIGABRT
```

| path | `terminate()` |
|---|---|
| every other model — `LocalWhisperSTT` ~1280 | `setTimeout(() => w.terminate(), 5000)` — **5s grace** |
| Nemotron — `sharedWorkerRegistry:303` | `capturedWorker.terminate()` — **immediate** |

The Nemotron branch explicitly opted out: *"No terminate()-after-a-timer here … there's nothing to defer."* There is — the worker thread can be inside a native `onnxruntime` `run()`, and killing it there aborts the process.

`stop()` only holds the worker open for pending **finals**; `shouldKeepWorkerForFinals` ignores `streamingTaskInFlight`. So an in-flight streaming task is orphaned **by design, for every model** — only the 5s grace made that survivable. Restored here.

The ONNX slot is still freed immediately rather than behind the timer, matching the non-Nemotron ordering, so a rapid meeting restart can't starve against the gate's cap of 2.

## 2. Why it produced nothing — no preload

`ModelPreloader` returned early for Nemotron, so the **~7s** load of its three ONNX sessions happened at *meeting start*. The consequences compound:

- VAD banks a backlog for those 7s, so the first dispatch is seconds of audio, not the 560ms streaming window — observed **4080ms** (mic) / **5670ms** (system)
- `streamingTaskInFlight` gates re-dispatch, so the whole streaming loop stalls behind that one oversized request
- any meeting shorter than backlog + inference ends before a single result returns

**This is precisely why Parakeet worked and Nemotron didn't** — Parakeet is warm from app launch and streams from the first second.

The early return had a real reason: Nemotron owns its worker through `sharedWorkerRegistry`, and layering this class's "one warm worker, hand it to whichever channel asks first" scheme on top would create a second, competing notion of who owns the worker. So this warms it **through the registry** instead — acquire a long-lived `preload-warm` channel and hold its release. The registry stays the single owner, and mic/system then JOIN a ready worker: `NemotronEngine.create()` reuses `nemotronSharedResources` and returns without loading anything.

Keeps every guarantee the normal preload path has — persisted recent-failure cooldown, free-memory refusal, workerPath existence check, and the crash sentinel (a native abort during session load kills the process before any catch runs; the sentinel is what makes the next launch treat the model as poisoned). The hold is dropped when the selected model changes away from Nemotron.

The warm hold occupies one ONNX slot for the app's lifetime. That is **not** new cost — it is exactly what `ModelPreloader` already does for every other model, and it matches the gate's documented shape ("Whisper + one background consumer" against the default cap of 2).

## 3. Observability (kept deliberately)

This path logged **nothing** on success, empty result, or hallucination-filtered result, and `sharedWorkerRegistry.ts` had no logging at all — "transcribed fine", "returned empty" and "never ran" were indistinguishable in a log. Four boundary logs now cover cold-start vs join, ONNX slot wait, dispatch size, and which of the three result outcomes occurred. They are what turned this diagnosis from guesswork into evidence, and the next person in this path will need them.

## A note on one source-assertion test

`LocalWhisperStuckWorker.test.mjs` asserts that `armStreamingWatchdog()` and `worker.postMessage` sit within **200 characters** of each other, as a proxy for "both on the same dispatch path". The dispatch log initially landed between them and broke that proximity check without changing behaviour. Moved the log above the arming rather than loosening a real invariant.

## Verification

- `typecheck:electron` — clean
- **670/670** across `electron/audio`, `electron/utils` and `electron/audio/whisper` — including the real-model go/no-go, the 18-language suite, `StartupModelValidation`'s preloader source assertions, and `OnnxSlotWeightedAcquisition`
- `Requires physical macOS verification` — that a meeting now transcribes end to end. Not yet observed in the running app.
- `Requires physical Windows verification` — preload and teardown timing are platform-neutral, but the cold load and the abort were only measured on macOS/CoreML.

## Known, unproven, not addressed here

`nemotronChains` is keyed per channel specifically so the two channels don't wait on each other, but both engines hold the **same three `InferenceSession` objects** via `getSharedResources()`. That permits concurrent `run()` on one CoreML-backed session. Not implicated by this evidence (a single mic dispatch and a single system dispatch, then an abort at teardown — not at overlap), so deliberately left out of this PR rather than bundled in on speculation.

https://claude.ai/code/session_011VSHtwCxXjMVdyheoF73Bd

---

# Final side-by-side verdict: Nemotron vs Parakeet, as shipped

Each model simulated under **its own in-app streaming policy** with real inference timed live (M-series, CPU EP, threads as shipped: Parakeet intra=1, Nemotron intra=4). 9.85s two-utterance audio:

| user-felt metric | Parakeet (in-app) | Nemotron (this branch) |
|---|---|---|
| first visible text | 4244ms | **1203ms** |
| incremental updates | 7 | **18** |
| inference per update | 66→565ms — **grows with segment** | **76–106ms constant** |
| final lag after speech ends | 565ms | **204ms** |
| transcript quality | "the the … brownown … lazzy" | clean, cased, punctuated |

Two structural reasons, both from this branch's fixes plus the models' shapes:

- Parakeet re-transcribes the whole growing segment every 1500ms (quadratic in segment length) and runs LocalAgreement-2 — on the short fixture its q8 artifacts ("the the / brownown") **never agreed**, so nothing appeared until the final. Nemotron's cache-aware delta path is constant-cost and emits directly per 560ms chunk.
- Raw RTF still favors Parakeet (0.058 vs 0.155 at shipped threads) — irrelevant to the user once both are far below realtime; profile shows Nemotron's remaining time is 81% encoder, so decoder-loop tricks were deliberately skipped (~4% for real risk).

Also fixed while benchmarking: the model re-emits **`<en-US>` language-tag tokens at utterance boundaries mid-stream** ("lazy dog. `<en-US>` The quick…"); both decode paths leaked them. Stripped exact-shape at the shared join point, with a negative test that non-tag bracket text survives. 673/673.

---

# Accuracy: 4.4% → 0.0% WER (Parakeet: 41.2% on identical audio)

User-reported inaccuracy, quantified on 4 real TTS sentences through the app decode path. The one systematic Nemotron error was a **dropped weak first word** at utterance start ("the meeting…" → "meeting…", "our quarterly…" → "Quarterly…", capitalized — the model believes the utterance starts there). Reproduces with no VAD in the loop, so it's model behavior at zero left context, not audio clipping.

Ruled out first: **chunk boundaries** (streamed decode is byte-identical to whole-audio on every sentence) and **precision** (no higher-precision ONNX export exists — the other HF repos are byte-identical mirrors or MLX/CoreML runtimes).

Fix: **50ms silence pre-roll at segment start, inside the engine**. 50ms and not 100ms because tr-TR — documented marginal at 0.0 overlap — flips chaotically with start offset (non-empty at 0/50ms, empty at 25/75/100ms), and 50ms is the measured point that fixes first words *and* keeps all 18 locales non-empty (tr-TR actually produced its best-yet output there). `NATIVELY_NEMOTRON_PREROLL_MS` overrides for future measurement. An arithmetic test proves exactly-once-per-segment application.

| WER, same 4 sentences | before | after |
|---|---|---|
| Nemotron (app path) | 4.4% | **0.0%** |
| Parakeet | 41.2% ("schedululed", "datababase", "teamam", one empty) | 41.2% |

674/674 including the 18-locale multilang suite.


evinjohnn:
@greptile-apps review

PR #484, opened 2026-08-18
Local profile intelligence, and fix the build without the premium submodule

StephenBiele:
## Summary

`premium/` is a private submodule, so profile intelligence (resume and job-description context) is absent from this tree rather than gated. This adds a local implementation behind the same contract, running on the user's own models.

**Nothing here bypasses, disables, or emulates a license check.** A new module resolver tries the premium module first and falls back to the local one only when premium is absent, so installing the private submodule restores today's behaviour with no code change. The `LicenseManager` call sites are deliberately not routed through it.

The first commit is independently useful and worth reviewing on its own.

### `fix(build)` — the source-available tree did not build

Two unguarded `require('../../premium/...')` calls in `resolveCompanySearchProvider.ts` failed the build on any checkout without the private submodule. esbuild runs with `bundle: true` and resolves literal specifiers at build time; a require inside a `try/catch` is left as a runtime call, but an unguarded one is fatal. So `npm run build:electron` failed on a clean clone, which failed `npm test` before a single test ran. Verified by stashing this work and rebuilding: those two requires were the only blocker.

Also repoints two type-only imports at premium paths, which were hard `TS2307`s in the same situation.

### `feat(profile)` — the local implementation

Five modules under `electron/localKnowledge/`. Each reuses existing engines rather than duplicating them:

- **DocumentReader** delegates PDF/DOCX/text parsing to `SafeDocumentTextExtractor`, adding normalization for embedding and error results instead of exceptions. Line endings and Unicode composition are normalized because the same resume otherwise chunks differently depending on the authoring platform.
- **profileNormalization** owns everything deterministic. The readers parse dates with a strict `/^\d{4}-\d{2}$/`, so a model writing `June 2024` would silently disable every tenure answer. Durations are never stored, since the existing code computes them at query time. A year-only range yields no date rather than an invented month.
- **ResumeExtractor** calls a local Ollama model in JSON mode at temperature 0, falling back to deterministic parsing when no model is available — following `OkfExtractor`'s principle that extraction must work when every provider is unavailable.
- **ProfileIndex** presents the profile to the existing `ModeHybridRetriever` under a reserved mode id instead of building a second RAG stack.
- **LocalKnowledgeOrchestrator** implements `processQuestion` against the existing deterministic evidence selector rather than inventing its own logic.

## Type of Change

- ✨ New Feature (local profile intelligence)
- 🐛 Bug Fix (standalone build)

## Testing & Environment

- [x] Manual test performed on: **macOS 15 (Apple Silicon)**, against a real Ollama install — `gemma4:12b-mlx` for extraction, `ollama:nomic-embed-text:768` for embeddings

| Check | Result |
| --- | --- |
| `npm run typecheck:electron` | clean |
| `npm run build:electron` | succeeds (previously failed on a clean tree) |
| `npm test` | 7112 pass, 63 fail — an **identical failing set to the pre-change baseline**; all 63 are in 40 files that `require` the absent premium submodule |
| New tests | 82 pass, 0 fail |

End to end on the real PDF fixture: ingest 2.9 s, index status `ready`, grounding returned in 17–25 ms against the caller's 2000 ms budget, `factualRecall` correctly true for skills and experience questions and false for a passage-only one.

## Notes for reviewers

Three decisions are load-bearing and easy to undo by accident:

1. **Retrieval is capped at 1200 ms**, under the caller's 2000 ms grounding budget (`IntelligenceEngine.ts:1636`). Overrunning that budget discards the whole result rather than degrading it, so the slow part is capped to guarantee the deterministic evidence returns.
2. **Deletion is written as direct SQL inside a transaction**, not delegated to `ModeHybridRetriever.removeFileIndex()`, which catches and logs its own failures. A call that never throws cannot roll a transaction back, and these rows carry resume content, so a partial delete is a privacy problem.
3. **No embedding pipeline is constructed here.** Per the note at `ModeContextRetriever.ts:1668-1680`, a pipeline nobody initializes keeps a null provider forever and silently degrades retrieval to lexical for the process's lifetime. The module waits for the shared pipeline instead.

Bugs found by running against a real model rather than stubs: reasoning models spent their whole budget thinking (one hit the 120 s timeout on a 335-character resume; `think: false` brought the same file to 4.7 s), and model auto-selection picked an installed embedding model that cannot answer a chat request.

## Known gaps

This is not user-visible yet, by design:

- Knowledge mode does not persist across restart.
- The upload remains gated in the renderer.
- JD structuring is not implemented, so the selector's `jd_*` answer shapes cannot fire. JD text is indexed and answerable from passages.

Happy to fold any of these in, or to split the build fix out as its own PR if that lands faster.


StephenBiele:
Closing — opened in error. This was not intended to go upstream.

PR #487, opened 2026-08-18
audit: full-app audit → fix autopilot (101 commits, zero regressions vs main)

evinjohnn:
Full-app audit → fix autopilot. 101 commits, 167 files, run one finding at a time: **confirm → reproduce → root-cause → fix → verify**, with every fix mutation-probed and every step diffed by failing test **name**, never count.

## Verification

Measured on the merged tree against a **freshly re-measured** baseline from current `main` (the earlier baseline went stale when main advanced 16 commits):

| | tests | pass | fail |
|---|---|---|---|
| `main` @ 893427e1 | 8051 | 7978 | 9 |
| this branch | 8139 | 8067 | 8 |

**Regressions: none** — the failing-name diff is empty in both directions. The two tests failing on main but passing here are the python local-runner cases, which nothing in this branch touches; that is environmental, not a fix claimed here.

`test:lib` 338/338 · `typecheck:electron`, `typecheck:ts7`, production build all clean.

## What this fixes

Beyond the earlier campaign findings, the last pass closed six static-review findings and then a second review pass over *those* fixes:

- **CR-03** — a non-`granted` Windows mic status was unactionable: `request-mic` reported success without doing anything off darwin, no settings link existed off darwin, and onboarding could never complete. Also found while confirming: the toaster painted the mic control **green over a denied device**.
- **CR-04** — the degraded-settings-store refusal was bypassed by two typed setters, so memory and disk diverged while the IPC reported success. Worse for `screenUnderstandingMode`, whose handler also **broadcast** the change to every window over an unchanged disk — on a setting whose copy promises "cloud vision is never called".
- **CR-05** — phone-mirror chat passed neither route flag, so it aborted 3s before the server's provider cascade could rescue the turn, and gave a cold-loading local model 7s instead of 30s.
- **CR-06** — a failed *data* repair permanently blocked a later *schema* migration (the vec0 cosine rebuild), leaving one database with mixed distance metrics read through a single threshold.
- **CR-07** — a finished STT session leaked a 5s keep-alive interval, permanently, once per session.
- **CR-08** — investigated and **dismissed**: optional chaining short-circuits the whole chain, so the reported crash cannot occur. No production change; the invariant is pinned so a refactor can't introduce it.

## Two things reviewers should know

**A fix in this branch was itself defective, and an independent review caught it.** CR-03 classified win32 `restricted` as "blocked by your organization" with a disabled button. Electron maps win32 `DeniedBySystem → "restricted"`, which is the *device-level* mic switch — the most common Windows denial, and exactly what the Settings panel fixes. That was worse than the state it replaced. Fixed in `60105df9`.

Notably, **the mutation probe could not have caught it**: the accompanying test asserted the wrong behaviour, so reverting the fix failed the test exactly as designed. Mutation testing proves a test is load-bearing, never that its premise is true.

**Migration numbering collided with `main` twice.** main claimed v28 for `meetings.user_titled` (a real `ALTER TABLE`) while this branch's page-count repair also sat on 28; whichever stamped first would have permanently suppressed the other. main keeps 28; this branch moved to 29/30, and the marker key is now version-free so it stops misdescribing itself.

## Still owed

**CR-03 requires physical Windows verification.** The corrected win32 branch has never run on Windows hardware — this machine reports `granted`, so `denied`/`restricted`/`not-determined` only ever ran under injected platform. Given that this exact branch already shipped one bad fix, it is worth exercising on real Windows before release.

Full reasoning, including every finding that turned out to be wrong or too narrow, is in `AUDIT_REPORT.md` §22–§24.


evinjohnn:
## Forward-merged `main` (219 commits) — was `CONFLICTING`, now `MERGEABLE`

### One conflict mattered far more than the other seven: a migration-version collision

`main`'s **v26 → v27 creates `usage_outbox`**. This branch's **v27 was the v22 page-count repair**. Both stamp `user_version = 27`, so whichever ran first would have **permanently suppressed the other**:

- a user coming from `main` would never get the page-count repair;
- a user from this branch would never get `usage_outbox`, breaking the usage ledger.

No textual merge can catch this — both sides are individually valid and neither conflicts *textually* on the version number. Resolved by letting the trunk keep its number:

| Version | Migration | Origin |
|---|---|---|
| 27 | `usage_outbox` | `main` |
| 28 | v22 page-count repair | this branch |
| 29 | vec0 cosine rebuild | this branch |

R-05's chain-stop (a swallowed failure must not let a later block stamp past it) survives the renumber. This branch was never released, so no installed profile can be on the old numbering.

### `main` still carries the credential data-loss bug

`main`'s `CredentialsManager` retains `fallbackMtime > keyringMtime → removeKeyringFile()`. That is the whole-profile-restore case (Time Machine, Migration Assistant, synced AppData) that **silently destroys the user's live API keys** — it is live on `main` today. This merge removes it. `main`'s newer work (provenance hashing, the re-key path, `reentryRequired`) is preserved; the file was rebuilt from `main`'s version with R-10 re-applied rather than untangling markers.

### Post-merge verification caught three defects in my own resolution

1. **A scope bug that would have shipped** — `let preferFallbackThisLoad` landed in `saveCredentials()` instead of `loadCredentials()` (`main` has three `keyringUnreadable = false;` sites; the anchored replace hit the first). Every credential load threw `ReferenceError`.
2. **A "smarter" R-10 that was unsound** — I first used `main`'s `fileIsOurs` to narrow the ambiguity (*fallback provably ours + keyring provably not ⇒ fallback authoritative*). Wrong for exactly the case R-10 protects: a whole-profile restore carries the **provenance record** along with the salt and both blobs, so the restored fallback hashes as ours while the user's current keyring hashes as foreign. The repro caught it writing `STALE-FROM-BACKUP` over live keys. mtime-newer is now always ambiguous; neither store is destroyed.
3. **Two repros pinned migration version numbers**, which legitimately changed here — re-anchored on migration purpose.

### Verification

| | tests | pass | fail |
|---|---|---|---|
| `main` (59bec00e) | 7673 | 7598 | 11 |
| this branch merged | 7725 | 7655 | 6 |

**Zero regressions against `main`** at name level. The only name that failed on the merge but not on `main` was `OkfPhase0FalseRefusalGuard`'s pin asserting the topic-blind evidence tier is OR'd into the repair gate — the owner-confirmed F-412 contract change — now updated and additionally pinned negatively.

All 12 repros pass on the merged tree; credential suites 66/66.

**Not claimed as fixes:** six `checkAnswerRelevance` corpus tests fail on `main` and pass here. Failure counts moved 12 → 6 between two merged runs with only one unrelated code change between them, so those model-dependent tests are flaky run-to-run. I'm reporting them, not taking credit for them.

**Windows:** still `Reviewed but not executed` locally. The CI Windows leg now runs — note this repo's Build Smoke Windows leg is `continue-on-error`, so a green tick there is not by itself evidence.

PR #493, opened 2026-08-21
audit follow-up: fix the 7 defects a review found in PR #487's own fixes

evinjohnn:
A high-effort review of #487 surfaced 8 findings in the campaign's own fixes. Seven were real; this branch fixes all seven. The eighth was refuted.

Each fix has a regression test that **fails on the unfixed source** — verified by reverting each file to its pre-fix state and re-running.

| Finding | Fix |
|---|---|
| R-16 | `LiveRAGIndexer`'s R-03 guards keyed on `meetingId`, but the only caller passes the constant `'live-meeting-current'` — so they could never fire. Now a per-session token. |
| R-17 | A desktop turn typed while a phone turn streamed was dropped in full: tokens rejected cross-surface, `done` unhonored, spinner stuck. The send path now claims the surface. |
| R-19 | `SettingsManager`'s read-failure path latched read-only with no quarantine — a permanent brick R-15's own docstring claims to have removed. |
| R-20 | A failed vector reap was swallowed, so R-12's transaction committed the meeting DELETE anyway, orphaning vectors. |
| R-21 | Re-stamping `embedding_space` left the old space's vectors in place, to be scored against new-space queries. Now clears, atomically. |
| R-22 | v29 lacked v28's `return`, re-arming R-05 for whoever adds v30. |
| R-23 | F-706 surfaced a Windows mic state nothing could resolve — `request-mic` is a no-op there that returned `true`. Now deep-links to the privacy pane. |

**Two corrections to the review**, both established empirically rather than argued:

- R-16's alleged mechanism (rowid reuse contaminating the next meeting's vectors) is **wrong** — `chunks.id` is `INTEGER PRIMARY KEY AUTOINCREMENT`, measured `[1,2,3] → [4,5]`, never reused. The finding survives on a worse mechanism proven by execution: the next meeting is silently killed.
- Finding #8 (an alleged `TypeError` in `AIProvidersSettings`) is **refuted** — optional chaining short-circuits the whole chain, `.then()` included. No bug, no change.

Also included: **R-24**, found while fixing R-19 — `set()` was `void`, so all 14 settings IPC handlers reported success *and broadcast their `*-changed` event* on a refused write, desynchronising every window onto a value disk never received.

## Validation

- typecheck (electron + renderer) clean
- `test:lib` 337/337 · db 24/24 · rag 274 pass / 5 pre-existing environmental · services 3197 tests, 3159 pass / 3 pre-existing environmental (same names as baseline)
- R-16 and R-17 additionally verified by **executing** the race, not just reading it
- R-21 covered behaviourally against the real compiled `VectorStore` and real SQLite, including the rollback path
- R-23's test drives **both** platform branches

## Not verified

- The renderer edits (`NativelyInterface.tsx`, the two onboarding components) are reviewed but **not executed** — the app was never launched.
- **Requires physical Windows verification.** R-23 makes the state resolvable; it does not confirm Windows behaviour on hardware.

https://claude.ai/code/session_01ESRcrBN7iFQ4F9uNgBh6V8

### DIFF
--- a/electron/db/DatabaseManager.ts
+++ b/electron/db/DatabaseManager.ts
@@ -3150,7 +3150,14 @@ export class DatabaseManager {
                     const ph = slice.map(() => '?').join(',');
                     try {
                         this.db!.prepare(`DELETE FROM ${table} WHERE ${column} IN (${ph})`).run(...slice);
-                    } catch (_) { /* dim table may not exist */ }
+                    } catch (e: any) {
+                        // R-20: getExistingVecDims() returns KNOWN_DIMS ∪ discovered, so on a
+                        // normal install most dims have no table and "no such table" is the
+                        // expected, harmless outcome. Swallowing EVERYTHING else too meant a
+                        // real reap failure was indistinguishable from that — the same silence
+                        // that let R-01 hide for a full campaign. Narrow it.
+                        if (!/no such table/i.test(String(e?.message ?? e))) throw e;
+                    }
                 }
             };
             for (const dim of dims) {
@@ -3158,7 +3165,14 @@ export class DatabaseManager {
                 if (summaryIds.length) deleteIn(`vec_summaries_${dim}`, 'summary_id', summaryIds);
             }
         } catch (e) {
-            console.warn(`[DatabaseManager] deleteVectorsForMeeting(${meetingId}) failed (non-fatal):`, e);
+            // R-20: R-12 wrapped this and the parent DELETE in one transaction so the
+            // meeting could not survive a failed reap. Swallowing here defeated that
+            // in the only direction that matters: the transaction saw no error and
+            // committed the DELETE anyway, leaving the meeting gone and its vectors
+            // orphaned in the vec0 tables — where they keep consuming KNN top-K slots
+            // that searchSimilarNative can no longer resolve back to `chunks`.
+            console.error(`[DatabaseManager] deleteVectorsForMeeting(${meetingId}) failed; aborting the delete:`, e);
+            throw e;
         }
     }
 
--- a/electron/db/DatabaseManager.ts
+++ b/electron/db/DatabaseManager.ts
@@ -1621,6 +1621,15 @@ export class DatabaseManager {
                 this.db.pragma('user_version = 29');
             } catch (e) {
                 console.error('[DatabaseManager] v29 vec0 cosine rebuild failed (leaving version at 28 to retry next launch):', e);
+                // R-22: stop here, exactly as v28 does. v29 is currently the LAST
+                // migration, so falling through is harmless TODAY — which is
+                // precisely why it would not stay harmless: the next migration
+                // added below would run and stamp user_version past a v29 that
+                // never applied, and `version < 29` would be false forever after.
+                // That is R-05 verbatim, and R-05 only became reachable because
+                // v28 was written this same way. Terminate the chain on failure
+                // instead of leaving the trap armed for whoever adds v30.
+                return;
             }
         }
 
--- a/electron/db/__tests__/MigrationChainTerminatesOnFailure2026_08_20.test.mjs
+++ b/electron/db/__tests__/MigrationChainTerminatesOnFailure2026_08_20.test.mjs
@@ -1,0 +1,67 @@
+// R-22 regression test.
+//
+// R-05: v28 fell through after a failed repair, so v29's block ran and stamped
+// user_version past a v28 that never applied — `version < 28` was false ever
+// after and the repair never retried. v28 was fixed to `return`.
+//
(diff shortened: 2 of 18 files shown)


evinjohnn:
## CI verified against a same-base baseline

The Windows leg of Build Smoke carries `continue-on-error: ${{ runner.os == 'Windows' }}` (build-smoke.yml:256), so its green tick is not evidence. Read the log instead, and compared it against run `32401843653` — the baseline on `3bd71d7a`, which is this branch's exact base.

| | baseline @ `3bd71d7a` | this branch |
|---|---|---|
| tests | 7954 | **7977** (+23) |
| pass | 7845 | **7868** (+23) |
| fail | 19 | **19** |

The 19 failures are the **identical set** by name — `comm -23` of the two sorted failure lists is empty. They are pre-existing and untouched by this branch: the `D:\D:\`-doubled bare-path imports in the CredentialsManager suite, an `ENOTEMPTY` on the reranker temp dir, and several Cropper/Lexical/Litellm/LocalEmbed files.

Every test added here **ran and passed on Windows**, including the R-23 platform-branch tests:

```
PASS: win32 no longer claims a request it cannot make succeeded
PASS: win32 mic step deep-links to the privacy pane instead
PASS: the ms-settings scheme is allowlisted for win32 only
PASS: 'unknown' is normalised rather than surfaced off-contract
PASS: a previous session's parked tick cannot write into the new session
PASS: a previous session's late stop() cannot tear down the new session
PASS: a failed re-stamp rolls back rather than half-clearing the meeting
PASS: no settings IPC handler reports success on a refused write
  … 23 in total
```

So these are **covered by automated Windows branch tests**, not merely reviewed — an upgrade on what the PR body claimed.

### Still not verified

`Requires physical Windows verification` stands for R-23. CI runs the tests, not the app's permission flow against a real privacy toggle.

`Reviewed but not executed` stands for the renderer edits. The entry-smoke job typechecks and **builds** the renderer on both platforms — so `NativelyInterface.tsx` and the onboarding components compile everywhere — but it never boots the app (`Verify packaged entrypoint exists` is an `accessSync`, not a launch). Meaningfully exercising R-17 needs a phone-mirror turn streaming concurrently with a desktop one.

evinjohnn:
Superseded — closing.

This PR targeted `audit/autopilot-2026-08-18`, which merged to `main` as #487 on 2026-08-22. The work has landed on `main` directly instead, as **`401407f4`**, rebuilt on top of the merged tree rather than merged as-is.

**Landed (5 of the 8 findings), each with a test that fails on the unfixed source:**

| | |
|---|---|
| R-17 | A desktop turn typed while a phone turn was streaming was swallowed whole — tokens rejected cross-surface, `done` unhonored. The send path now claims the surface before the round-trip. |
| R-20 | A failed vector reap no longer lets R-12's transaction commit the meeting DELETE and orphan the vectors. |
| R-21 | Re-stamping `embedding_space` now clears the old space's vectors first, in one transaction, and runs ahead of the store loop. |
| R-22 | v30 (renumbered from v29 when #487 landed) no longer falls through on failure, re-arming R-05 for whoever adds v31. |
| R-16b | The post-await embedding writes are guarded, so a parked batch cannot write vec0 rows for chunks F-411 already purged. |

**Dropped as superseded** — checked against what landed first rather than assumed:

- **R-16** — CR-02 fixes it, same design, found independently.
- **R-23** — CR-03 is more complete: a shared `src/lib/micPermissionPolicy` so main and renderer cannot disagree, and `classifyMicStatus` already treats `'unknown'` as usable with a better-researched rationale (Electron's win32 path fails *open* — `GetActivationFactory`/`CreateFromDeviceClass` failures return Allowed). I also reverted my `ms-settings:` allowlist widening after seeing `permissions:open-mic-settings` call `shell.openExternal` directly: it would have been added attack surface for a path nothing uses.
- **R-19 — I had this wrong.** `SettingsDegradedStoreMutators2026_08_22` asserts the opposite contract: on a read failure, stay degraded and leave the user's file **untouched**. That is safer than my quarantine, which would rename a healthy `settings.json` on a transient `EMFILE`. Its test caught it.

**R-24** was reduced to what was not already done: `set()` already reported whether the write landed, but twelve handlers ignored it and broadcast their `*-changed` event anyway, putting every window on a value disk never received.

**Finding #8 stays refuted** — optional chaining short-circuits the whole chain, `.then()` included. No bug, no change.

Still open and unaddressed, from the earlier CI comparison: `Requires physical Windows verification` no longer applies (R-23 dropped), but the renderer edits in R-17 remain **reviewed and loaded, not exercised** — meaningfully testing it needs a phone-mirror turn streaming concurrently with a desktop one.

evinjohnn:
Superseded by 401407f4 on main — see the breakdown above.

PR #495, opened 2026-08-21
Fix NVIDIA NIM audio delivery and interview auto-answer

mohakmalviya:
## Summary

- Deliver live PCM frames to NVIDIA NIM with the correct protobuf field.
- Improve quiet Windows microphone speech capture.
- Trigger the answer planner after interviewer speech ends.
- Reconnect NVIDIA NIM streaming sessions after transient disconnects with bounded buffering.

## Validation

- Native Windows audio module rebuilt successfully.
- Electron bundle built successfully.
- Renderer build completed successfully.

This is the follow-up to the NVIDIA NIM speech integration and is separate from the cloud-provider PR.


evinjohnn:
Superseded by #497, merged to `main` as 3832e6f4. Closing this one — thanks @mohakmalviya, the diagnosis here was the valuable part and it was right.

**What carried over.** `handleSuggestionTrigger` — the method `IntelligenceEngine` documents as *"the primary auto-trigger path"* — genuinely had no production caller. The only call site was the `__e2e__:ask` harness, and the speculative prefetch that runs on interviewer partials never reaches the UI by design, so it only warmed a cache nothing consumed. Automatic answers were dead. That's now fixed, behind a **Settings > General → Auto Answer** toggle (default OFF, so the hotkey stays the only path unless you opt in).

**What changed in the rework.**

- **Trigger moved from `speech_ended` to the final interviewer transcript.** `getLastInterviewerTurn()` only ever returns a final turn, and for the REST providers `speech_ended` is what *starts* the upload — so the 250 ms debounce read the previous question by construction there. Firing stale is worse than not firing: the trigger Jaccard-compares against the in-flight speculative run, rejects, and cancels the correctly-prefetched answer before generating one for the previous question.
- **Added a mode + cooldown gate.** `runWhatShouldISay` opens by aborting any live stream with `'superseded'`, so an auto-trigger could kill an answer the user had just requested by hand.
- **`NvidiaNimStreamingSTT.ts` dropped entirely** — #492 had already landed all of it in a stronger form. The `closing` latch was also a regression: `finalize()` set it, only `start()` cleared it, and `finalizeMicSTT()` runs mid-meeting on every manual-answer press, so the first press would have killed mic transcription for the rest of the session.
- **Mic VAD scoped to Windows** rather than disabled everywhere, so macOS keeps the stage that rejects typing, fans and speaker bleed. The `speech_threshold_rms` 100 → 50 edit was dropped as inert — it's only the *initial* `adaptive_threshold`, which the first non-speech frame overwrites with ~59.
- **Brand mark** — `main` landed its own (lobehub `nvidia-color.svg`) while this was open.

CI ran for the first time on #497 (this PR was `CONFLICTING`, so its workflows never queued): macOS 8093 pass / 0 fail, Windows green on the enforcing suites with all 11 new tests passing there.

PR #512, opened 2026-08-26
fix(windows): Interface Opacity preview no longer paints a black window

evinjohnn:
## What broke

Settings → Interface Opacity, slider held, is the one moment the launcher window is *supposed* to be see-through: the renderer hides every DOM layer and `setLauncherOpacityPreview()` strips the native `backgroundColor`, so the desktop shows behind the mockup.

The launcher rounded-corner work in this same branch put an opaque black backing on `#root` to fill the splash→launcher gap. That backing sits **above** the native layer the preview strips and **below** every DOM layer the preview hides, so neither half of the preview reached it. Holding the slider painted a black rectangle.

This did **not** undo the 2.8.5 fix (`4ed6c39e`, "make launcher opacity preview work on Windows/Linux") — all four of its halves are still in place and still correct. A new opaque layer was added *above* the one that fix punches through, so it was outflanked rather than reverted.

## Verification

`tests/css/launcher-corner-radius.check.mjs` gains the preview state (win32 dark, win32 light, linux): see-through interior, no hairline ring, plus a string check that `index.html`'s early-paint mirror carries the same guard (that copy is not `!important`, so it is exactly what wins once the stylesheet's copy stops matching).

Three mutations were run, each failing with the right message:

| Mutation | Result |
|---|---|
| Drop the `#root` guard | interior `rgb(0,0,0)` — reproduces the reported bug |
| Drop the ring guard | trips all three configs (the light case justifies the `!important`) |
| Guard win32 but not linux in `index.html` | caught |

That harness calls `disableHardwareAcceleration()`, so it proves the CSS cascade but **cannot** see a GPU-side black box. A separate probe ran under real GPU compositing, sampling actual screen pixels with a magenta window behind a transparent one:

```
GPU compositing: enabled | scaleFactor=1.25 | Electron 43.1.0
✓ bare-transparent     see-through  rgb(255, 0, 255)
✓ preview-fixed        see-through  rgb(255, 0, 255)
✓ preview-absent       black        rgb(0, 0, 0)
✓ preview-no-contain   see-through  rgb(255, 0, 255)
```

`contain: paint` was suspected of forcing an opaque composited layer and is ruled out.

## Commits

| | |
|---|---|
| `10fb04e6` | `chore(gitignore)` — ignore local AI-assistant/editor configs |
| `efd6b163` | `feat(windows)` — rounded frameless launcher + launcher/overlay swap choreography |
| `5f201651` | **`fix(windows)` — the Interface Opacity preview fix** |
| `8775bfa7` | `fix(windows)` — stop showing macOS permission walkthroughs on Windows |
| `2ce1ce0c` | `fix(stealth)` — Windows shortcut guard becomes opt-out; keep win32 typed |
| `26e97822` | `chore(repo)` — track OCR screen fixtures + graph instruction doc |

The corner-rounding commit deliberately lands *before* the guard so the regression is bisectable. The launcher-corner harness ships with the guard commit, so no commit fails its own test.

## Cross-platform

- **Windows / Linux** — preview reveals the desktop; the black backing and ring return on release. Splash→launcher gap still opaque, corners unchanged.
- **macOS** — untouched by the opacity fix; none of these selectors ever matched there and the attribute is inert.

**Validation:** `Covered by automated Windows branch tests` · `Covered by automated Linux branch tests` · `Tested physically on Windows` (the opacity preview) · `Reviewed but not executed on macOS`.

**Commands run:** `npm run test:css:launcher-corners`, `npx tsc -p tsconfig.json --noEmit`, plus the GPU probe above.

## Remaining risks

- The swap choreography's on-screen timing and the `ms-settings:` mic deep link still want physical Windows verification.
- `2ce1ce0c` flips the WH_KEYBOARD_LL shortcut guard to **on by default**, so the hook is now installed for every Windows user rather than only opt-ins. The Rust compiles but is not exercised here — it needs `npm run build:native` and an AV/EDR check on real Windows.
- This branch also carries `6ceb2b93`, a pre-existing local commit that was never pushed. Local `main` was 1 ahead / 20 behind `origin/main` when the branch was cut, so this is **not** rebased onto current `origin/main`.

https://claude.ai/code/session_01K9YNdVhkUKsxnY6HLh1pbd


evinjohnn:
-

PR #533, opened 2026-09-01
Update development server configuration to use IPv4 loopback address. 

Power-Howdy:
## Summary
Changed references from 'localhost' to '127.0.0.1' in package.json, Vite config, and multiple Electron window helper files to ensure compatibility across platforms. Additionally, modified build script to handle premium Electron submodule presence more gracefully.

Fixes # 
This issue was detected while setting up development environment on my own, and I created PR without an issue.

## Type of Change
- [x] 🐛 Bug Fix
- [ ] ✨ New Feature
- [ ] 📝 Documentation
- [ ] ⚡ Performance Improvement
- [ ] 🛠️ Chore / Tooling

## Testing & Environment
- [x] Manual test performed on: **[e.g. macOS, Windows 11]**

### Verification Steps

1. 

## Checklist
- [x] My code follows the code style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new errors or warnings
- [x] Relevant tests have been performed / updated

### DIFF
--- a/electron/CropperWindowHelper.ts
+++ b/electron/CropperWindowHelper.ts
@@ -4,7 +4,7 @@ import path from "node:path"
 const isDev = process.env.NODE_ENV === "development"
 
 const startUrl = isDev
-    ? "http://localhost:5180"
+    ? "http://127.0.0.1:5180"
     : `file://${path.join(app.getAppPath(), "dist/index.html")}`
 
 /**
--- a/electron/ModelSelectorWindowHelper.ts
+++ b/electron/ModelSelectorWindowHelper.ts
@@ -5,7 +5,7 @@ import { attachNoActivate } from "./utils/windowsFocusPolicy"
 const isDev = process.env.NODE_ENV === "development"
 
 const startUrl = isDev
-    ? "http://localhost:5180"
+    ? "http://127.0.0.1:5180"
     : `file://${path.join(app.getAppPath(), "dist/index.html")}`
 
 import type { WindowHelper } from "./WindowHelper"
--- a/electron/SettingsWindowHelper.ts
+++ b/electron/SettingsWindowHelper.ts
@@ -6,7 +6,7 @@ import { attachNoActivate } from "./utils/windowsFocusPolicy"
 const isDev = process.env.NODE_ENV === "development"
 
 const startUrl = isDev
-    ? "http://localhost:5180"
+    ? "http://127.0.0.1:5180"
     : `file://${path.join(app.getAppPath(), "dist/index.html")}`
 
 type WindowActivationOptions = {
--- a/electron/WindowHelper.ts
+++ b/electron/WindowHelper.ts
@@ -46,7 +46,7 @@ function traceOverlayResize(event: string, data: Record<string, unknown>): void 
 }
 
 const startUrl = isDev
-  ? 'http://localhost:5180'
+  ? 'http://127.0.0.1:5180'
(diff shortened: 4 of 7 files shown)


evinjohnn:
Thanks for this — the diagnosis is correct and the loopback half is now on `main` as aee4ecfe, with authorship preserved to you.

I split the PR and took only the `localhost` → `127.0.0.1` change. Reasoning below so it's not a black box.

## What was merged

Your root cause checks out, and it isn't Windows-only — I reproduced it on macOS. Same tree, only `vite.config.mts` swapped:

| config | listening socket | `127.0.0.1` | `[::1]` |
|---|---|---|---|
| `main` (no `host` → default `localhost`) | `IPv6 TCP [::1]:5188 (LISTEN)` | **refused** | 200 |
| this PR (`host: '127.0.0.1'`) | `IPv4 TCP 127.0.0.1:5187 (LISTEN)` | **200** | refused |

Vite binds IPv6-only, and it only works today on machines where the client happens to agree on `::1`. Where `wait-on` resolves IPv4 first — Node 17+ uses verbatim DNS order — `npm start` hangs and never launches Electron, exactly as you described. Pinning both ends to one literal address is the right fix, and your coverage was complete: after the change there is no `localhost:5180` left in any code path.

## What was not merged: `scripts/build-electron.js`

This one changes release-build integrity, and I'd like to keep it out of a dev-config fix.

```js
const CORE_SMOKE = process.env.NATIVELY_CORE_SMOKE === '1' || !premiumPresent;
```

Same premium-less tree, only the build script swapped:

| build script | exit | output |
|---|---|---|
| `main` | **1** | `✘ Could not resolve "../../premium/electron/knowledge/TavilySearchProvider"` (+5 more) |
| this PR | **0** | `premium submodule missing; compiling with private imports external` |

The bundle it emits carries ~30 `require("../../premium/…")` calls that can never resolve, and every call site swallows them (`main.ts:1301`, `ipcHandlers.ts:401` are `try { … } catch { /* not available */ }`). So the failure mode is: build succeeds, app boots, Pro features are silently dead — no signal at build time or runtime.

CI is safe (`release-macos.yml:89` does `test -f premium/electron/knowledge/CompanyResearchEngine.ts` under `set -euo pipefail`), but local `npm run app:build:signed` has no equivalent guard, and there's no Windows release workflow at all — so every Windows package is cut through the unguarded path. That's the case I want to keep failing loudly.

It also runs against an invariant stated in `build-smoke.yml:64-67`: *"a missing trusted secret is an explicit failure, never a silent downgrade."* Worth noting the existing test named **"core smoke externalizes private runtime imports only behind an explicit opt-in"** still passes on your branch — its regex is an unanchored substring match, so `|| !premiumPresent` slips through. That's a gap in our test, not something you did wrong.

## If you want to pursue the ergonomics fix

It's a real papercut — a fresh clone without the private submodule can't run `npm run build:electron`, and that's worth solving. A separate PR would be welcome if it:

1. anchors the assertion in `build-smoke-fork-safety.test.mjs` (`/^const CORE_SMOKE = …=== '1';$/m`) and adds coverage for the new implicit path;
2. keeps `app:build` / `app:build:signed` failing hard when premium is absent — ideally reusing the `test -f` guard from `release-macos.yml` so Windows gets it too;
3. leaves `build:electron` alone and instead surfaces a clear message telling contributors to use `npm run build:electron:core-smoke`.

Closing this one since the fix it's named after has landed. Thanks again — the IPv6 diagnosis was a good catch and it was affecting more platforms than the title claimed.

PR #591, opened 2026-09-20
fix: scope keybinds (#517), refresh Codex models (#573), retain coding context (#539), and voice RAG follow-ups (#552)

### LINKED ISSUE #517: [Bug]:  Default global shortcuts suppress common browser keyboard events system-wide
### What happened?

Natively registers numerous OS-global shortcuts, including common browser and
input chords such as:

- `Ctrl+B`
- `Ctrl+R`
- `Ctrl+Y`
- `Ctrl+Enter`
- `Ctrl+1` through `Ctrl+7`
- `Ctrl+Shift+Arrow`

The defaults are defined in `electron/services/KeybindManager.ts:21-68` and
registered through Electron `globalShortcut` at lines 335-365.

Because the OS-global handler consumes the completing chord, the focused browser
may not receive its normal keyboard event or action. This conflicts with browser
shortcuts and creates a stable missing-event pattern that an instrumented page
can observe when the user presses one of the configured chords.

### Also tell us, what did you expect to happen?

The default configuration should minimize persistent global registrations,
avoid common browser chords, or register task-specific shortcuts only during an
active session. Users should also have an obvious one-click option to disable all
global shortcuts.

### Steps to Reproduce

1. Fully exit Natively, including its tray process. Hiding its window is not
   sufficient because global shortcuts remain registered while the process runs.
2. Open `https://example.com/` in Chrome.
3. Press `F12` or `Ctrl+Shift+I` and select the **Console** tab.
4. Paste this listener and press `Enter`:

   ```js
   for (const type of ["keydown", "keyup"]) {
     window.addEventListener(type, event => {
       console.log(
         type,
         event.code,
         `ctrl=${event.ctrlKey}`,
         `shift=${event.shiftKey}`,
         `alt=${event.altKey}`
       );
     }, true);
   }
   ```

5. Click once in the webpage so the page, rather than DevTools, has keyboard
   focus.
6. With Natively still closed, press `Ctrl+Shift+Right Arrow` once.
7. Confirm that the Console records `ArrowRight` keyboard events.
8. Start Natively V2.8.8 and wait for its launcher to finish loading.
9. Return to `https://example.com/` and click the page again so Chrome is
   foreground.
10. Press `Ctrl+Shift+Right Arrow` once more.
11. Observe that Natively moves to the right because the chord is mapped to
    `window:move-right`, while the page no longer receives the same completing
    `ArrowRight` event sequence.
12. As a second user-visible comparison, enable **Preserve log** in DevTools and
    press `Ctrl+R`:
    - With Natively closed, Chrome reloads `example.com`.
    - With Natively running and the shortcut registered, Natively handles
      **Reset / Cancel** and the browser may not perform its normal reload.
13. Fully exit Natively and repeat steps 5-7 to confirm normal events return.

### Environment

- OS: Windows 11 x64, NT 10.0 build 26200
- Source tag: `V2.8.8`
- Commit: `a2485e49613269bc6bce3f2bf43eca50321bbb09`
- Root `package.json` version: `2.8.7`
- Electron: `43.1.0`
- Chrome: `151.0.7922.174`

### Relevant Log Output

```shell
[KeybindManager] Registered global shortcut: CommandOrControl+Shift+Right -> window:move-right


No error is emitted; this is the successful shortcut-registration path.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### LINKED ISSUE #539: [Bug]:  Natively loses active coding-problem context and returns unrelated/general answers
## What happened?

On Natively 2.8.8 for Windows, **What to Answer** and short manual follow-ups can lose the active coding problem, route the turn as non-coding, and return an unrelated or contradictory answer.

This happened during a technical interview about two microservices that encrypt/decrypt messages, rotate the active key at the top of each hour, and retain old keys for decryption. The observed failures were:

| Input or spoken follow-up | Actual response | Expected response |
|---|---|---|
| `show me how you would implement in python` | Unrelated dynamic-programming `count_ways(n)` / stairs solution | Python implementation of the active rotate/encrypt/decrypt problem |
| `show in python` | The same unrelated `count_ways(n)` solution | Continue the active coding problem in Python |
| `show the solution in python` | The same unrelated `count_ways(n)` solution | Continue the active coding problem in Python |
| `write a function for rotate, encrpt, descrypt` | Character-shifting/Caesar-cipher code | Key rotation plus real whole-message symmetric encryption/decryption |
| Interviewer clarified “encrypt the entire text, not character by character” | Natively again suggested iterating through characters with `ord()` / `chr()` | Use a whole-message cipher/API (for example authenticated symmetric encryption) and preserve the key ID/version needed for decryption |

The attached local telemetry shows that this was not a provider timeout:

- 11 What-to-Answer presses occurred during the coding section.
- All 11 were recorded as `isCoding: false`.
- Only 2/11 selected `prior_assistant_responses`; 9/11 excluded them.
- All 11 completed and rendered in 2.643–5.136 seconds (median 3.396 seconds).
- There were no provider timeouts or crashes for these presses.

The session therefore demonstrates a context-selection/routing failure, not merely poor model output.

### Audit findings

I reproduced the routing behavior directly against the `AnswerPlanner.js` bundled in the installed 2.8.8 app. Results were the same in both General and Technical Interview mode and for both manual and What-to-Answer sources:

```text
"show in python"                                      -> general_meeting_answer
"show the solution in python"                         -> general_meeting_answer
"how to encrypt entire text"                          -> general_meeting_answer
"How would you encrypt the entire text instead of
 doing it character by character?"                    -> general_meeting_answer
```

The source audit found several compounding causes:

1. The live WTA path reads a hard-coded **180-second** context window, then sparsifies it to at most **12 turns**. The original problem statement was already outside this window by the first logged WTA press. See [`SessionTracker.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/SessionTracker.ts#L124-L131), [`IntelligenceEngine.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/IntelligenceEngine.ts#L1399-L1451), and [`transcriptCleaner.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/transcriptCleaner.ts#L157-L221).
2. The transcript extractor always chooses the newest non-greeting interviewer fragment, even when it is only a constraint or comment. It may lower confidence but does not fall back to the preceding answerable coding question. See [`transcriptQuestionExtractor.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/transcriptQuestionExtractor.ts#L364-L403) and its [answerability check](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/transcriptQuestionExtractor.ts#L456-L478).
3. `SessionTracker` already accumulates multi-segment coding questions, but manual WTA does not use the detected coding question to rescue routing. See [`SessionTracker.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/SessionTracker.ts#L517-L530) versus the authoritative question passed to the planner in [`IntelligenceEngine.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/IntelligenceEngine.ts#L1779-L1786).
4. Elliptical implementation follow-ups lack the planner's required coding keywords and fall into `general_meeting_answer`. Only a recognized `follow_up_answer` receives prior assistant responses; other routes fail closed for that layer. See [`AnswerPlanner.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/AnswerPlanner.ts#L1969-L1993), its [context-layer rules](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/AnswerPlanner.ts#L1013-L1047), and [`contextRoute.ts`](https://github.com/Natively-AI-assistant/natively-cluely-ai-assistant/blob/218c8756bb4c1a462d6428e03f9a95f576d34a19/electron/llm/contextRoute.ts#L116-L133).

The transcript also contained duplicated cross-channel speech/echo (approximately 28% of utterances touched a near-duplicate pair). That makes the small turn budget fail faster, but it does not explain why explicit Python/encryption follow-ups are classified as general answers.

## Steps to Reproduce

1. Start a meeting in Natively 2.8.8 on Windows. The routing replay reproduces in either General or Technical Interview mode.
2. Have the interviewer state a coding problem across several speech segments, for example:
   - Service A encrypts and Service B decrypts.
   - Rotate the active key at the top of every hour.
   - Retain at least five key versions so in-flight/older messages remain decryptable.
3. Continue the conversation for more than three minutes so the original statement leaves the hot transcript window.
4. Press **What to Answer** after a short follow-up such as “show the solution in Python,” or enter that phrase manually.
5. Observe that the answer is routed as general/non-coding, lacks the full problem, or uses an unrelated prior coding problem.
6. Clarify: “How would you encrypt the entire text instead of doing it character by character?”
7. Observe that the planner again selects `general_meeting_answer` and may repeat per-character shift code instead of answering the clarified cryptography question.

An additional deterministic split-turn case is:

```text
INTERVIEWER: Implement an LRU cache with get and put.
INTERVIEWER: You can assume capacity is positive.
[Press What to Answer]
```

The extractor selects only the latest constraint fragment; the previously accumulated coding question is not used by manual WTA.

## Expected behavior

- Maintain an explicit, session-scoped “active coding problem” independently of the 180-second display/transcript window.
- Join constraint/clarification fragments to that active problem instead of replacing it with the latest fragment.
- Treat “show [the solution] in Python,” “implement this,” and “encrypt the whole text” as coding continuations when the active topic is coding.
- Route the turn consistently through the coding contract and include the necessary prior problem/context.
- Prefer a clarification over inventing or retrieving an unrelated problem when context cannot be resolved.
- Deduplicate echoed cross-channel transcript segments before applying the 12-turn budget.

## Environment

- OS: Windows x64, NT `10.0.26200.0`
- Natively executable FileVersion: `2.8.8`
- Natively executable ProductVersion: `2.8.8.0`
- Installed `app.asar` timestamp: `2026-08-28T09:05:16-07:00`
- Installed `app.asar` SHA-256: `31269B562CF3ECF4D15187266BCB970D8BF3AA0FD5161B04E424EC5A4A2FCAC0`
- `verboseLogging`: enabled
- Configured primary action mode: `brainstorm`, but affected telemetry events explicitly report `source=what_to_answer`, so this setting did not reroute those presses.
- Direct Assist: not present in this installed build/settings, so it was not the cause.

There is a version-identification problem in the repository: tag `V2.8.8` points to commit `a2485e49`, where `package.json` still says `2.8.7`, while current `main` is 59 commits later and still reports `2.8.8`. Please expose the exact build commit in the app and identify which commit produced the hash above.

## Relevant Log Output

```text
Sanitized session summary:
coding-section WTA presses:          11
planned as coding:                    0 / 11
planned as non-coding:               11 / 11
prior assistant context selected:     2 / 11
prior assistant context excluded:     9 / 11
completed/rendered:                  11 / 11
provider timeouts:                    0
latency range:                        2.643–5.136 s
median latency:                       3.396 s

Observed wrong answer signature:
def count_ways(n): ...

Expected topic:
rotating-key encrypt/decrypt implementation for two services
```

The raw transcript, database, settings, and telemetry contain personal interview/profile data and identifiers, so they should not be posted publicly. I can provide a narrowly redacted trace privately if a maintainer needs it.

## Suggested regression coverage

- Preserve an active coding problem for longer than 180 seconds and across many fragmented STT turns.
- Latest interviewer turn is a constraint, with the preceding turn being the coding ask.
- Coding continuations: `show in python`, `show the solution in python`, `implement this`, `encrypt the entire text`.
- Duplex transcript contains near-duplicate `ME` and `INTERVIEWER` segments.
- No screen/DOM context is available; the app must still use the active transcript problem and must not reuse stale cross-session context.

## Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### LINKED ISSUE #552: [Bug]: JIT RAG pre-flight drops conversation history, so follow-up questions lose their referent
## Summary

While a meeting is being live-indexed, the JIT RAG pre-flight intercepts every typed chat message and answers it from transcript chunks alone. The conversation history never reaches the model, so follow-up questions that depend on the previous turn are answered without it.

The first question of a session is fine. The follow-up is not.

## Reproduction

1. Start a meeting so JIT indexing is active and has embedded chunks.
2. Ask a self-contained question in chat, e.g. `lc 573`. Correct answer.
3. Ask a follow-up with no standalone content: `do via stack`, or `explain that more`.

Expected: the follow-up is understood as referring to the previous answer.
Actual: it reaches the model as a bare fragment with no referent, and is answered against whatever transcript chunks happen to embed near it.

Outside a meeting (or before any chunks are indexed) follow-ups work correctly, which makes this look intermittent.

## Cause

`handleManualSubmit` builds the conversation context and then runs the RAG pre-flight *before* using it. A successful RAG result returns early, so the history-carrying path never runs:

`src/components/NativelyInterface.tsx` (line numbers from current `main`)

```js
const conversationContextForSubmit = buildConversationContextFromMessages(messages);   // ~7467

const ragResult = await window.electronAPI.ragQueryLive?.(userText || '');             // 7472
if (ragResult?.success) {
  return;                                                                              // 7475
}

// ...only reached when RAG declines:
await window.electronAPI.streamGeminiChat(userText, imagePaths, conversationContextForSubmit);
```

No layer below could compensate — none of them accept history:

| Layer | Signature |
| --- | --- |
| `preload.ts` | `ragQueryLive: (query: string)` |
| `ipcHandlers.ts:13139` | `'rag:query-live', { query }: { query: string }` |
| `RAGManager.ts:280` | `queryMeeting(meetingId, query, abortSignal?)` |
| `prompts.ts:80` | `buildRAGPrompt(query, context, scope, intent)` |

The same pattern exists at `NativelyInterface.tsx:7249` in the voice path (`handleAnswerNow`).

## Why it is easy to miss

It never errors. RAG returns a fluent, confident answer, so it reads as the assistant being unhelpful rather than as a bug.

There is also a partial self-heal that hides the worst case: when retrieval finds nothing, `RAGManager` throws `NO_RELEVANT_CONTEXT_FOUND`, the handler returns `{fallback:true}`, and regular chat answers *with* full history. So it silently recovers exactly when retrieval fails hardest, and silently misfires when retrieval half-matches — which is the case with no escape hatch.

Contentless follow-ups are also the worst input for embedding retrieval, since they carry no content words to embed.

## Questions before anyone spends time on this

I may be missing context, so rather than assume a direction:

1. **Is this intended?** `conversationContextForSubmit` is computed and then used ten lines later on the fallback path, which reads like an oversight to me — but if meeting-scoped RAG is deliberately meant to answer only from the transcript, then the real issue is the gate, not the missing history.

2. **Which fix would you prefer?** Threading an optional `conversationHistory` through the four layers is the smallest change. But `MEETING_RAG_SYSTEM_PROMPT` says *"Answer questions based ONLY on the provided meeting excerpt"*, so history would be visible and off-limits unless that rule is scoped to claims about the meeting. That prompt change seems like the part you would want to own.

3. **Should the gate narrow instead, or as well?** The pre-flight currently asks whether transcript data is *available*:

   ```js
   if (!ragManager.isLiveIndexingActive('live-meeting-current') || !ragManager.hasLiveChunks()) {
     return { fallback: true };
   }
   ```

   Not whether the question is *about* the transcript. A similarity floor in addition to the existing `chunks.length === 0` check would route weak matches through the `{fallback:true}` path that already works. Would you want that instead of, or alongside, passing history?

4. **Is `RAGRetriever.detectScope()` intended for this?** It looks unsuitable to me — it ends with `return currentMeetingId ? 'meeting' : 'global'`, which is the same availability-equals-relevance assumption. Flagging in case someone reaches for it.

Happy to put together a patch for whichever direction you prefer, or to leave it as a report if this is already on your radar.

## Environment

- Version 2.8.8, macOS (Apple Silicon)
- Reproduced with a Gemini key; the path is provider-independent

### LINKED ISSUE #573: Codex model picker goes stale: dead presets, no source warning, no provider refresh
## Problem

Installed Natively 2.8.8 (macOS arm64, up to date per auto-updater) with a ChatGPT (Codex) subscription OAuth provider shows only **ChatGPT 5.5** as the latest selectable model — and the picker still lists the long-retired `gpt-5.3-codex-spark`, `gpt-5.3-codex`, and `gpt-5.4-mini` alongside it. That is the crux: every one of those three is backend-rejected for a ChatGPT account ("The '<id>' model is not supported when using Codex with a ChatGPT account"), so the GUI is offering models that 400 on selection while hiding every current provider model above 5.5. My persisted `settings.json` even has `codexCliFastModel: gpt-5.3-codex-spark`, a dead default.

The provider actually serves 5 list-visible models today (from a fresh `~/.codex/models_cache.json`, Codex CLI 0.154.0, fetched 2026-09-17):

- `gpt-6-astra` (pri 1)
- `gpt-5.6-sol` (pri 4)
- `gpt-5.6-terra` (pri 7)
- `gpt-5.6-luna` (pri 8)
- `gpt-5.5` (pri 12)

Inspecting the 2.8.8 `app.asar`: `CODEX_MODEL_REASONING_SETS` caps at `gpt-5.5` / `gpt-5.5-codex` — zero references to `gpt-5.6` or `gpt-6`. The picker ceiling is hardcoded, so new provider models can never appear until a Natively release catches up.

Two compounding UX issues found in source (`main`, v2.9.0):

1. **No CLI-missing warning.** `AIProvidersSettings.tsx` shows "Model list from your Codex CLI" when a catalogue exists, but when `readCodexModelCatalog()` returns `unavailable` (CLI never installed/never run — the common case, since Natively signs into ChatGPT itself per #558) the picker silently falls back to built-in presets with no indication the list may be stale. A user with no CLI has no way to know they are seeing presets, not the provider list. (I originally installed Natively with no CLI installed and had no idea.)
2. **No provider refresh.** The catalogue is read once from `models_cache.json`; Natively never re-fetches from the live backend even though it holds its own OAuth bearer token. The CLI refreshes that file on its own schedule, and the HTTP-direct `CodexCliService` never spawns the CLI, so the file can go arbitrarily stale (plus a Finder-launched packaged app does not inherit a shell-exported `CODEX_HOME`).

## Expected

- When the model list is unavailable or stale, the GUI should say so and direct the user to reconnect ChatGPT. It must not silently substitute a hardcoded roster.
- Codex CLI must not be required. Do not fetch on general app launch or before Codex authentication. On each authenticated AI Providers page mount/reload, show the current in-memory list immediately and refresh asynchronously from the provider. Other model pickers should remain cache-only.

## Note (to preempt a follow-up)

"Fast"/"light" variants are not separate model slugs in this account's provider response. Speed is model metadata (`service_tiers`, currently including `priority` where available), and the **Fast Mode Model** picker draws from the same provider model list. The old `gpt-5.3-codex-spark` default is retired and backend-rejected for ChatGPT accounts.

## Environment

- Natively 2.8.8 (auto-updater: no update available), macOS arm64
- Codex CLI 0.154.0, `codex login` (ChatGPT account), fresh `models_cache.json`
- Natively log shows `Codex CLI enabled with model: gpt-5.5`, settings pinned to `codexCliModel: gpt-5.5` / stale `codexCliFastModel: gpt-5.3-codex-spark`

markdudov:
## Overview
This PR addresses and fixes four critical issues across global keybindings, Codex model synchronization, multi-turn coding problem retention, and live meeting voice follow-ups:

- **Fixes #517:** Global shortcuts suppress common browser & IDE keyboard events system-wide.
  - Scoped `general:reset-cancel` to in-app (`isGlobal: false`), preventing OS-wide hijacking of browser reload (`Ctrl+R`).
  - Restricted window navigation/movement hotkeys (`window:move-*`) so they never register globally in launcher mode, eliminating interference with word-level text selection (`Ctrl+Shift+Arrow`).
  - Added a master toggle `globalShortcutsEnabled` in `KeybindManager.ts` (persisted across sessions and exposed via IPC and preload) and surfaced a toggle switch with informational status in `SettingsOverlay.tsx` (Keybinds tab).

- **Fixes #573:** Codex model picker goes stale / dead presets (`gpt-5.3-codex-spark`, `gpt-5.3-codex`), missing catalog warning & refresh.
  - Updated `CODEX_CLI_MODEL_PRESETS` in `modelUtils.ts` to active verified models: `gpt-6-astra`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, and `gpt-5.5`.
  - Updated `CODEX_MODEL_REASONING_SETS` in `CodexCliService.ts` and `AIProvidersSettings.tsx` to include `gpt-6` and `gpt-5.6` model families while removing retired `spark` and `5.3-codex`.
  - Added an in-place notice banner and "Refresh Models" button in `AIProvidersSettings.tsx` when the local CLI catalogue is absent or needs manual sync.

- **Fixes #539:** Natively loses active coding context across window switches / split turns (>180s transcript window), ignores consecutive interviewer constraints, squanders turn budget on duplex mic/speaker echo.
  - Expanded `isCodingContinuation` and `BARE_CODE_TOKENS` in `codingFollowup.ts` to recognize language indicators (`python`, `cpp`, `java`, `typescript`, etc.) and direct action phrases (`implement this`, `encrypt the whole text`).
  - Added contiguous interviewer turn accumulation in `transcriptQuestionExtractor.ts` so multi-sentence prompts and constraints split across pauses are merged into `latestQuestion`.
  - Implemented `deduplicateEchoedSegments` in `transcriptCleaner.ts` to drop acoustic echo bleed between mic and speaker before spending the turn budget.
  - Retained `activeCodingProblem` with timestamp in `SessionTracker.ts` and spliced the original problem into `preparedTranscript` and `priorInterviewer` in `IntelligenceEngine.ts` when rolling transcript eviction (>180s) would otherwise discard it.

- **Fixes #552:** Spoken follow-up turns in live meetings drop chat conversation context.
  - Guarded JIT RAG pre-flight (`ragQueryLive`) in `handleAnswerNow` (`NativelyInterface.tsx`) when prior chat history exists and the spoken input is an anaphoric follow-up, routing through the conversational prompt path and ensuring `conversationContextForSubmit` is preserved.

## Verification
- Automated unit test suites added and passing:
  - `electron/services/__tests__/GlobalShortcutsScoping.test.mjs` (5/5 pass)
  - `electron/services/__tests__/CodexModelPresetsRefresh.test.mjs` (4/4 pass)
  - `electron/llm/__tests__/SplitTurnCodingExtractor.test.mjs` (4/4 pass)
  - `electron/intelligence/__tests__/ActiveCodingProblemRetention.test.mjs` (3/3 pass)
  - `src/lib/__tests__/ragVoiceFollowupContext.test.mjs` (1/1 pass)
- Total: 17/17 tests passing.


evinjohnn:
Thanks @markdudov. Most of this PR is now on `main`, merged in `ceb08021` with you as co-author. Some parts were narrowed along the way, so here is what landed and what didn't.

**Landed**

- **#517 shortcuts** (`0284acd7`): Reset / Cancel is no longer global, `window:move-*` is not registered in launcher mode, and there is a new "Global shortcuts" switch in Settings → Keybinds. Two changes from your version:
  - The switch is stored in `settings.json`, so `keybinds.json` stays an array and older builds keep custom bindings.
  - When it's off, Toggle Visibility stays global. Without it, a hidden stealth window has no way back.
- **#573 Codex** (`0a7a6404`, `0bf86359`): gpt-5.6 and gpt-6 reasoning-effort entries. Before this, `gpt-5.6-terra` matched bare `gpt-5`, so an xhigh pick was silently sent as low. GPT-6 Astra and GPT-5.6 Sol are added to the built-in list, with ChatGPT 5.5 kept first because it's the default.
- **#539 coding context** (`bb6c3f39`):
  - "show in python", "show the solution in python", "show me how you would implement in python" and "implement this" are treated as coding continuations.
  - What to Answer puts the detected coding problem back in the prompt once it has left the 180s window.
  - A constraint turn ("You can assume capacity is positive.") is joined to the ask right before it.

**Left out, and why**

- **The contiguous interviewer-turn join:** it joined every contiguous turn, so a 25-sentence monologue became a 1,654-character "question", and 3 existing extractor tests broke. The version on `main` joins one turn, within 30s, and only when that turn is a pure constraint.
- **The language tokens in `isCodingContinuation` and the lower `looksLikeCodingQuestion` threshold:** "Tell me about your experience with Python" and "Have you worked in go?" became coding follow-ups, and the old coding problem was spliced into those answers.
- **`deduplicateEchoedSegments`:** it dropped the candidate's own restatements ("so you want me to reverse a linked list in place, right").
- **The #552 voice-path change:** `main` had already fixed #552 on the typed-chat path the issue describes. The voice-path guard also skipped meeting retrieval for any short question containing "it/that/why".
- **The amber "catalogue unavailable" banner:** most users have no Codex CLI, so it would always show, and its text was unreadable in the light theme.
- **The source-grep tests:** replaced with tests that run the compiled code. They still passed with the regressions above in place.

**Verification:** tested live on macOS with a real model. For the #539 scenario, `main` answered with unrelated code and `codingTask=false`; this branch answered the actual key-rotation problem in Python. Shortcut registration and the Settings switch were also checked in the running app. Windows was not tested physically. The Windows hook's chord table follows the same switch, and a test covers that.

Closing, since the useful parts are merged.

